PR #17, opened 2021-06-18
feat: multiline entry

MarvinJanssen:
Adds the ability to write Clarity over multiple lines. It counts open braces and will hint the last brace type you are in. The feature makes it easier to write Clarity interactively as you are not forced to either (a) write the whole function definition in a single line, or (b) write code in a different editor and then paste it into the REPL. If you close the braces in the wrong order, it discards the input and allows you to recover (See last example).

Looks like this (every line was followed by a line-feed, notice the prompt changes):

```
clarity-repl v0.12.2
Enter "::help" for usage hints.
Connected to a transient in-memory database.

>> (+ 1 2)
3
>> (+
(.. 1
(.. 2
(.. 3
(.. )
6
>> (define-public (example)
(.. (ok true)
(.. )
→ .ST000000000000000000002AMW42H.contract-2 contract successfully stored. Use (contract-call? ...) for invoking the public functions:
>> {
{.. key: "val"
{.. }
{key: "val"}
>> (+
(.. }
Error: expected closing ), got }
(.. 1 2)
3
```

I know `rustyline` has a `Validator` trait but it did not seem to give me the flexibility I desired (like changing the prompt.)

### DIFF
--- a/history.txt
+++ b/history.txt
@@ -1,3 +1,0 @@
-#V2
-::get_contracts
-(+ 2 2)
--- a/src/frontend/terminal.rs
+++ b/src/frontend/terminal.rs
@@ -7,6 +7,25 @@ use std::io::{stdin, stdout, Write};
 
 const VERSION: Option<&'static str> = option_env!("CARGO_PKG_VERSION");
 
+fn complete_input(str: &str) -> Result<Option<char>, (char, char)> {
+    let mut brackets = vec![];
+    for character in str.chars() {
+        match character {
+            '(' | '{' => brackets.push(character),
+            ')' | '}' => match (brackets.pop(), character) {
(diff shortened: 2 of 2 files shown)


friedger:
LGTM

lgalabru:
Thank you @MarvinJanssen! 
This approach looks good to me. How does the terminal react when you paste a multiline snippet?

MarvinJanssen:
No issues here. In fact, the history also functions properly in both cases and gives you the whole snippet.

What do you think of the prompt? I looked at sqlite3 and it looks like this:

```
sqlite> select '
   ...> a';
```

I like showing the open brace in the OP though.

lgalabru (approved):
LGTM, thank you @MarvinJanssen!

PR #24, opened 2021-09-22
feat: toggle_costs to show costs automatically

fariedt:
I would like to see the costs of calls I make without having to type `::get_costs` before each one.  This adds a toggle to prepend that to every input typed into the cli.


lgalabru (approved):
LGTM, thank you for your contributions @fariedt!

PR #26, opened 2021-09-22
feat: load history file on cli startup, if present

fariedt:
This makes clarinet's console a lot more useful: load a previous session's history file if it's present.


lgalabru (approved):
LGTM, thank you @faried!

PR #38, opened 2021-11-30
Add check-checker (and supporting AST visitor)

obycode:
Remaining todo items:
- [x] Properly printing out warning diagnostics
- [x] Refine warnings


obycode:
Example:
```
$ cat counter.clar
(define-public (tainted (amount uint))
   (stx-transfer? amount (as-contract tx-sender) tx-sender)
)
$ clarinet check 
test:2:19: warning: use of potentially tainted data
   (stx-transfer? amount (as-contract tx-sender) tx-sender)
                  ^~~~~~
test:1:26: note: source of taint here
(define-public (tainted (amount uint))
                         ^~~~~~
```

obycode:
- [ ] Decide what to do about the "contract successfully stored" and other messages that are now being printed out but were not previously

src/analysis/ast_visitor.rs
```diff
@@ -0,0 +1,985 @@
...
+                        // ),
+                        // AsContract => self.traverse_as_contract(&args[0]),
+                        // AtBlock => self.traverse_at_block(&args[0], &args[1]),
+                        // GetBlockInfo => {
+                        //     self.traverse_get_block_info(args[0].match_atom().unwrap(), &args[1])
+                        // }
+                        // ConsError => self.traverse_err(&args[0]),
+                        // ConsOkay => self.traverse_ok(&args[0]),
+                        // ConsSome => self.traverse_some(&args[0]),
+                        // DefaultTo => self.traverse_default_to(&args[0], &args[1]),
```
lgalabru, line 210: Should we do something with this cases?

obycode: Oh yes. All of the cases that weren't needed for the check-checker didn't get implemented yet. I'll add these in.

src/repl/interpreter.rs
```diff
@@ -100,21 +128,68 @@ impl ClarityInterpreter {
...
+        for (n, line) in lines.enumerate() {
+            if let Some(comment) = line.trim().strip_prefix(";;") {
+                if let Some(annotation_string) = comment.trim().strip_prefix("#[") {
+                    let span = Span {
+                        start_line: (n + 1) as u32,
+                        start_column: (line.find('#').unwrap_or(0) + 1) as u32,
+                        end_line: (n + 1) as u32,
+                        end_column: line.len() as u32,
+                    };
+                    if let Some(annotation_string) = annotation_string.strip_suffix("]") {
```
lgalabru, line 154: Should we `trim_end()` here?

src/analysis/ast_visitor.rs
```diff
@@ -0,0 +1,985 @@
...
+        &mut self,
+        expr: &'a SymbolicExpression,
+        list: &'a [SymbolicExpression],
+    ) -> bool {
+        let mut rv = true;
+        if let Some((function_name, args)) = list.split_first() {
+            if let Some(function_name) = function_name.match_atom() {
+                if let Some(define_function) = DefineFunctions::lookup_by_name(function_name) {
+                    rv = match define_function {
+                        // TODO: I think all of these can be simplified to just use unwrap, since the structure has already been checked at this point.
```
lgalabru, line 39: Correct, if this pass is running after the type checker, then you have the guaranty that we are looking at a contract correctly formatted.

lgalabru: That being said, the thing I like with this approach, is that it could be a vehicle for future work, helping us working around the constraint "1 error raised per analysis". In the context of the stacks blockchain, there's no reason to continue parsing the contract if it has at least one error. In the context of developer tooling, we don't have to follow the same rules. What do you think?

obycode: That depends on how good of a job we can do on working around errors to maintain an AST that is close enough to meaningfully continue.

src/analysis/check_checker/mod.rs
```diff
@@ -0,0 +1,2097 @@
...
+            AtomValue(value) => self.visit_atom_value(expr, value),
+            Atom(name) => self.visit_atom(expr, name),
+            List(exprs) => self.traverse_list(expr, &exprs),
+            LiteralValue(value) => self.visit_literal_value(expr, value),
+            Field(field) => self.visit_field(expr, field),
+            TraitReference(name, trait_def) => self.visit_trait_reference(expr, name, trait_def),
+        }
+    }
+
+    fn traverse_define_public(
```
lgalabru, line 203: Unless there is some sort of cascade already in place, I think we should treat `traverse_define_private` in a similar fashion: a public-function could delegate its writing operations to a private-function.

src/analysis/ast_visitor.rs
lgalabru: We would probably abort any child nodes checks, however we could maybe continue with the siblings / parents nodes?

obycode: But yes, I agree. We should report as many errors as possible in one pass.

src/analysis/ast_visitor.rs
obycode: Added traversers for all in 89c7611

src/analysis/ast_visitor.rs
obycode: Thinking about this some more now. I think that the check for a properly formed AST has to happen before this AST visitor, and in the case that we want to continue processing past an error, we'll need to insert a placeholder that maintains a well-formed AST, otherwise, it is too complex to handle in all later passes. That means that the unwrap technique should be fine here.

src/analysis/check_checker/mod.rs
obycode: When the public function is called, all of the parameters are added to the taint sources. Then, all expressions within that public function are checked. That means if a private function is called, it will already be properly checked. I'll add a unit test to ensure this.

src/analysis/ast_visitor.rs
obycode: Addressed in ed03067

src/analysis/check_checker/mod.rs
obycode: Oh, but we do need to transfer the taint onto the parameters, similar to the way `let` is handled. I will add that.

src/analysis/ast_visitor.rs
obycode: Force-pushed over that commit, it is now in 44e4abb

src/analysis/check_checker/mod.rs
obycode: This is actually a bit more complex than I initially thought, to get the most developer-friendly behavior. Here is what I am thinking. @lgalabru, let me know if you have any thoughts.
1. Require all checking to be done in the public function, i.e. only checked values can be passed to a call to a private function - this keeps the checking close to the source, which is desirable for readability
2. Add an annotation to allow a specific parameter of a private function to be unchecked - this allows for the use case of a private function which is a filter for unchecked data (dirty in, clean out)
3. If unchecked data is allowed (via 2), the return value must be a checked value

src/repl/interpreter.rs
lgalabru: yes!

src/analysis/check_checker/mod.rs
obycode: 604066c adds 1 and 2, except that the annotation applies to all parameters for now. Later, we can enhance this to specify which parameters are allowed to be unchecked. It also does not currently check the return value.

src/repl/interpreter.rs
obycode: Fixed in 604066c

src/analysis/check_checker/mod.rs
obycode: The return value checking requires some more significant changes, so I will include that in a separate commit.

lgalabru (approved):
LGTM, thank you for taking on this feature and addressing my comments!
Could you please create an issue for listing the remaining todos that will be addressed in a subsequent PR, so we have them in our radar?
Thanks!

src/repl/session.rs
```diff
@@ -287,7 +291,8 @@ impl Session {
                     true,
                     Some("Deployment".into()),
                 ) {
-                    Ok((_, result)) => {
+                    Ok((ref mut res_output, result)) => {
+                        output_err.append(res_output);
```
lgalabru, line 295: On `Ok(_)`, I think we should be appending in `output`, instead of `output_err`?

lgalabru: Following up on the extra log that we would like to remove in clarinet, I think we could move the snippet L:545 in this file

```rust
if let Some((ref contract_name, _, _, _, _)) = result.contract {
    let snippet = format!("→ .{} contract successfully stored. Use (contract-call? ...) for invoking the public functions:", contract_name.clone());
    output.push(green!(snippet));
}
```

to line 522:
```rust
let mut result = match self.formatted_interpretation(
    snippet.to_string(),
    None,
    true,
    None,
) {
    Ok((mut output, result)) => {
        if let Some((ref contract_name, _, _, _, _)) = result.contract {
            let snippet = format!("→ .{} contract successfully stored. Use (contract-call? ...) for invoking the public functions:", contract_name.clone());
            output.push(green!(snippet));
        }
        output
    },
    Err(result) => result,
};
```

So that the extra line is only added when developers are interacting with the contracts from the console.

obycode: If we append to output, then the warnings are suppressed in addition to that "contract successfully stored" message. The problem is that both the warnings and that message are together coming out of `formatted_interpretation`. There is only one `Vet<String>` returned from that method, because before, there was only diagnostics if it failed. Should I just add another buffer to that tuple being returned, so we have an output and an err_output coming out of `formatted_interpretation`?

obycode:
I just rebased on top of latest develop branch.

lgalabru:
Ship it 🚀

PR #39, opened 2021-12-03
fix: set costs_version

reedrosenbluth:
set `costs_version` config option

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -101,6 +101,9 @@ jobs:
           RUSTFLAGS: "-Zinstrument-coverage"
           LLVM_PROFILE_FILE: "clarity-repl-%p-%m.profraw"
         run: cargo build && cargo test
+      
+      - name: Functional Test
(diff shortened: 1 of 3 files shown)


lgalabru (approved):
LGTM, thank you @reedrosenbluth! How digressing would it be to improve the `clarity-repl` cli, so that the bin can be invoked with a clarity operation - as in:
```bash
$ clarity-repl (+ 1 1)
```
If we had that, we could add a simple test to the CI, and fail if the command returns a non-zero status code. @CharlieC3 did [something similar for clarinet](https://github.com/hirosystems/clarinet/blob/a62ffebcba6d1c6b541f92a7cc39a37a88a95710/.github/workflows/ci.yaml#L106), it would let us make sure that the binary that we're shipping is working as expected.

reedrosenbluth:
@lgalabru not at all, working on this now!

reedrosenbluth:
@lgalabru is this what you had in mind?

lgalabru:
Awesome, yes, yes, yes! Thank you!

PR #44, opened 2021-12-16
fix: proposal for filtering extra logs

lgalabru:


obycode (approved):
LGTM

obycode:
I'll fix the errors after merging it in. The problem is just that the line count is off because of this line of output being removed.

PR #48, opened 2021-12-16
feat: check argument count to user-defined funcs

### LINKED ISSUE #47: Check for invalid argument count in calls
Aaron reported that the VM does not currently report an error when checking a contract with a call to a user function with the wrong argument count, however, this will cause a runtime error. It will eventually be fixed in the VM, but in the meantime, we can add a simple analysis pass using the new AST visitor framework to detect this and report an error from `clarinet check`.

obycode:
For example, for the following contract:
```clarity
(define-private (foo (amount uint))
    (ok amount)
)

(define-public (main)
    (ok (foo u1 u2))
)
```

This analysis pass will generate the error:
```
test:6:9: error: incorrect number of arguments in call to foo (expected 1 got 2)
    (ok (foo u1 u2))
        ^~~~~~~~~~~
```

Resolves #47

### DIFF
--- a/src/analysis/call_checker.rs
+++ b/src/analysis/call_checker.rs
@@ -1,0 +1,266 @@
+use crate::analysis::annotation::Annotation;
+use crate::analysis::ast_visitor::{traverse, ASTVisitor, TypedVar};
+use crate::analysis::{AnalysisPass, AnalysisResult};
+use crate::clarity::analysis::analysis_db::AnalysisDatabase;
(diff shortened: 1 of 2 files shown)


lgalabru (approved):
LGTM! How would you feel about adding a quick unit test? :)

obycode:
> LGTM! How would you feel about adding a quick unit test? :)

Oh yeah. Definitely.

lgalabru (approved):
LGTM, thank you @obycode!

PR #50, opened 2021-12-16
chore: update crates + repo metadata

lgalabru:


Cargo.toml
```diff
@@ -12,8 +12,8 @@ exclude = [
     ".husky",
     ".git*"
     ]
-homepage = "https://lgalabru.github.io/clarity-repl/"
-repository = "https://github.com/lgalabru/clarity-repl"
+homepage = "https://hirosystems.github.io/clarity-repl/"
```
obycode, line 15: Is it easy to update this to a later version? I saw that it is currently running 0.11.0.

obycode (approved):
LGTM

Cargo.toml
lgalabru: That'd be great - it's in our radar with https://github.com/hirosystems/clarity-repl/issues/36

PR #60, opened 2022-01-05
feat(check-checker): allow private function filter

obycode:
This changeset allows a user-defined private function to act as a filter
in the caller -- meaning that tainted data passed into the function can
be treated as filtered after the call. This is a common usage pattern
and should remove many false positive warnings in real smart contracts.


lgalabru (approved):
LGTM, thank you @obycode!

PR #63, opened 2022-01-05
ci: fix workflow quotes

CharlieC3:
For some reason it seems the paths-ignore rule is not being considered upon releases, which causes the release process to abruptly end by a new workflow instance that cancels the running one. All of the GH docs show to use single quotes, so it may help, although I'm not positive.


obycode (approved):
LGTM

PR #70, opened 2022-01-07
updated link to hirosystems repository

criadoperez:
The index.html was also not refering to the hirosystems repository. Changed it.


obycode (approved):
LGTM. Thanks @criadoperez!

obycode:
After #75 is merged, this can be rebased and then the CI jobs should all pass.

obycode:
I opened #76 with a cleaned up commit log so that we don't need to merge this one with two merge commits.

PR #73, opened 2022-01-10
feat: add 'filter' annotation

### LINKED ISSUE #72: Add filter annotation to check-checker
In smart contracts, there will sometimes be a check that the caller of the function is a trusted principal, then after that, the inputs from that caller should be trusted, as discussed in items 1 and 2 in #62. For example:

```clarity
(define-public (set-owner (address principal))
  (begin
    (asserts! (is-eq tx-sender (var-get owner)) (err ERR-NOT-AUTHORIZED))
    (ok (var-set owner address))
  )
)
```

The check-checker could support a new annotation to handle this case:
```clarity
(define-public (set-owner (address principal))
  (begin
    ;; #[filter(*)]
    (asserts! (is-eq tx-sender (var-get owner)) (err ERR-NOT-AUTHORIZED))
    (ok (var-set owner address))
  )
)
```

The argument passed to the `filter` annotation specifies a parameter which should be considered filtered after the following expression, or a `*` to specify that all parameters should be considered filtered.

obycode:
The filter annotation allows developers to tell the check-checker to
filter one or more inputs on an expression that cannot be detected as a
filter. This is useful when a check on some unrelated inputs indirectly
validates another input. For example:

```clarity
(define-public (withdraw (amount uint))
    (begin
        ;; #[filter(amount)]
        (asserts! (is-eq tx-sender (var-get admin)) (err u400))
        (stx-transfer? amount (as-contract tx-sender) tx-sender)
    )
)
```

The filter annotation takes either a comma-separated list of parameter
names to filter, or a '*', which filters everything.

Resolves #72

### DIFF
--- a/src/analysis/annotation.rs
+++ b/src/analysis/annotation.rs
@@ -1,9 +1,13 @@
 use crate::clarity::representations::Span;
+use crate::clarity::ClarityName;
 use regex::Regex;
+use std::convert::TryFrom;
 
 #[derive(Debug)]
 pub enum AnnotationKind {
     Allow(WarningKind),
+    Filter(Vec<ClarityName>),
+    FilterAll,
 }
 
 impl std::str::FromStr for AnnotationKind {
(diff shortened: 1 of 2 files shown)


lgalabru (approved):
Looks great, thank you for the extensive testing!

PR #75, opened 2022-01-11
fix: resolve CI failure for forks

obycode:
The GH_TOKEN secret is required for the semantic release step, but that
only runs on a PR to main. Since forks can open a PR to develop, which
needs to run the build and testing, but not the semantic release, we can
use the GITHUB_TOKEN instead. Fix from @CharlieC3.


lgalabru (approved):
👍

PR #85, opened 2022-01-27
Updated parser

obycode:
I am still doing some testing, but it is in good enough shape to start a review.


src/repl/ast/parser/error.rs
```diff
@@ -0,0 +1,76 @@
...
+            IllegalTraitName => "illegal trait name".to_string(),
+            InvalidPrincipalLiteral => "invalid principal literal".to_string(),
+            InvalidBuffer => "invalid hex-string literal".to_string(),
+            NameTooLong(name) => format!("illegal name (too long), '{}'", name),
+            UnexpectedToken(token) => format!("unexpected '{}'", token),
+            ExpectedClosing(token) => format!("expected closing '{}'", token),
+            TupleColonExpected => "expected ':' after key in tuple".to_string(),
+            TupleCommaExpected => "expected ',' separating key-value pairs in tuple".to_string(),
+            TupleValueExpected => "expected value expression for tuple".to_string(),
+            IllegalClarityName(name) => format!("illegal clarity name, '{}'", name),
```
lgalabru, line 57: Do you think we could revisit the level for this class of issues, and go with a `Warning` or `Info`?
With Clarinet, developers can add `links` to their Clarinet manifest files, which pull the source of the contracts from mainnet / testnet nodes. If we go with an `error`, that will be breaking these projects.

src/repl/ast/parser/mod.rs
```diff
@@ -0,0 +1,2384 @@
...
+            | Token::Greater
+            | Token::GreaterEqual => {
+                let name = ClarityName::try_from(format!("{}", token.token)).unwrap();
+                let mut e = PreSymbolicExpression::atom(name);
+                e.span = token.span;
+                Some(e)
+            }
+            // TODO: For now, comments are being treated as whitespace. In the future,
+            //       they should be attached to the following expression
+            // Token::Comment(comment) => {
```
lgalabru, line 611: We could handle the comments and collect these mini-ast in a side structure, if you wanted.

src/repl/interpreter.rs
```diff
@@ -70,23 +70,44 @@ impl ClarityInterpreter {
         contract_identifier: QualifiedContractIdentifier,
         cost_track: bool,
         coverage_reporter: Option<TestCoverageReport>,
-    ) -> Result<ExecutionResult, (String, Option<Diagnostic>, Option<Error>)> {
-        let mut ast = self.build_ast(contract_identifier.clone(), snippet.clone())?;
-        let (annotations, mut diagnostics) = self.collect_annotations(&ast, &snippet);
+        parser_version: u32,
+    ) -> Result<ExecutionResult, Vec<Diagnostic>> {
+        let (mut ast, mut diagnostics, success) =
+            self.build_ast(contract_identifier.clone(), snippet.clone(), parser_version);
```
lgalabru, line 76: If we're not adding the "fail fast" mode in this PR (used in future blockchain environment), could you open a PR for it? Thanks!

src/repl/session.rs
```diff
@@ -630,7 +634,7 @@ impl Session {
         name: Option<String>,
         cost_track: bool,
         test_name: Option<String>,
-    ) -> Result<ExecutionResult, (String, Option<Diagnostic>, Option<Error>)> {
+    ) -> Result<ExecutionResult, Vec<Diagnostic>> {
```
lgalabru, line 637: Thank you for cleaning up these!

lgalabru (commented):
This improvements are super exciting, and the implementation is really elegant, thank you @obycode!
I added a few feedbacks, curious to get your thoughts!

src/repl/ast/parser/error.rs
obycode: Can you give an example of what that looks like? It would be nice if we could still catch and report an error for a truly illegal name and also handle that case.

src/repl/ast/parser/mod.rs
obycode: Yes, I like that. Are you okay with that going into a separate PR after the initial set of changes in merged in? I'll open an issue for that.

obycode: Opened #88.

src/repl/interpreter.rs
obycode: Sure, I opened #89 for this. It could be very simple to include in this PR, but I'd like to discuss how that should be managed, so let's do it later.

src/repl/session.rs
obycode: 👍

src/repl/ast/parser/error.rs
obycode: I realized that there is one ugly change needed in order to support the weird names that are currently legal. Currently, the lexer just looks for identifiers, without any context. If the `>` character is allowed in an identifier, then the lexer will need to have context about whether it is looking for a trait-identifier, in which case the `>` might be closing the trait identifier, or any other identifier, in which case that is just part of the name. 

The v1 parser is not able to report an error when defining a trait with a bad name like this, but only gets confused when it is used:
```clarity
(define-trait multipl>ier
    ((multiply (uint uint) (response uint uint)))
)
(define-public (my-multiply (untrusted <multipl>ier>) (a uint) (b uint))
    (contract-call? untrusted multiply a b)
)
```
Old parser output:
```
$ clarinet check
error: Expected whitespace or a close parens. Found: '>'
```

I can solve this pretty easily, but I don't love it.

obycode: @lgalabru, please take a look at f20e58c with this modification.

lgalabru (approved):
LGTM, thank you @obycode!

obycode:
The only change in that last push was just rebasing on the latest develop branch (I guess that wasn't the right way to do it). Can you approve again? Thanks!

lgalabru (approved):
👍

PR #86, opened 2022-01-27
fix: add checks for argument counts to map-* funcs

obycode:
The type-checker in the clarity library checks for the case where there
are too few arguments, but not too many. This may be fixed in the
type-checker in 2.1 (because it is consensus breaking, see
stacks-network/stacks-blockchain#3018), but for now, we can do this in
the REPL.

Resolves: hirosystems/clarinet#228


lgalabru (approved):
LGTM, thank you @obycode!

PR #91, opened 2022-02-08
Improvements to the check-checker and configuration

obycode:


src/analysis/check_checker/mod.rs
```diff
@@ -70,12 +133,14 @@ impl<'a, 'b> CheckChecker<'a, 'b> {
...
         annotations: &'a Vec<Annotation>,
+        settings: Settings,
     ) -> CheckChecker<'a, 'b> {
         Self {
             db,
+            settings,
             taint_sources: HashMap::new(),
             tainted_nodes: HashMap::new(),
-            diagnostics: Vec::new(),
+            diagnostics: HashMap::new(),
```
lgalabru, line 143: Should we go with a `BTreeMap`, to make the traversals more handy when using the diagnostics?

lgalabru (approved):
Thrilled by this enhancements, thank you @obycode!

src/analysis/check_checker/mod.rs
obycode: We don't actually care about the order by the key though. The usage is only for inserts, removes, and then ultimately flattening it to a vector (below). Would a BTreeMap be more efficient for that flattening part?

```rust
        // Collect all of the vecs of diagnostics into a vector
        let mut diagnostics: Vec<Vec<Diagnostic>> = self.diagnostics.into_values().collect();
        // Order the sets by the span of the error (the first diagnostic)
        diagnostics.sort_by(|a, b| a[0].spans[0].cmp(&b[0].spans[0]));
        // Then flatten into one vector
        Ok(diagnostics.into_iter().flatten().collect())
```

obycode:
Sorry, @lgalabru, your PR requires a manual merge on this one so I need you to approve once more.

obycode:
> Sorry, @lgalabru, your PR required a manual merge on this one so I need you to approve once more.

I also changed the defaults as discussed.

lgalabru (approved):
👍

PR #103, opened 2022-02-21
fix: report an error for CRLF line-endings

obycode:
Only LF is supported. The parser will report an error on the first
instance of a CRLF (not on every line).


src/repl/ast/parser/lexer/error.rs
```diff
@@ -55,6 +56,7 @@ impl DiagnosableError for LexerError {
             SingleSemiColon => "unexpected single ';' (comments begin with \";;\"".to_string(),
             UnknownSymbol(c) => format!("unknown symbol, '{}'", c),
             NoteToMatchThis(c) => format!("to match this '{}'", c),
+            UnsupportedLineEnding => "unsupported line-ending '\\r', only '\\n' is supported".to_string(),
```
lgalabru, line 59: Should we also be explicitly using LF / CLRF denominations to improve the quality of the hints? Or add another message in `suggestion()`?

obycode: Good point, some people may understand CRLF/LF but not \r\n. Does the suggestion take the form of a patch? One problem with that could be that this would be a very large patch, since it would change every line. I haven't actually looked into how the suggestion gets displayed since we haven't used those yet. I will investigate.

The other problem could be that the user's editor might be saving with CRLF automatically, so even after they apply the suggestion, it comes back. Maybe an error and a note?

> error: unsupported line-ending '\r', only '\n' is supported
> note: you may need to change your editor from CRLF mode to LF mode

lgalabru: Adding a note sounds good to me!

lgalabru (approved):
LGTM!

PR #107, opened 2022-02-24
fix: consider as-contract in check-checker

obycode:
Operations on the sender's own wallet are considered safe, but this
is not correct if the operation is inside of an `as-contract`.


obycode:
I thought I had PRed this one last week. Almost forgot about it 😬!

lgalabru (approved):
LGTM!

PR #108, opened 2022-02-25
fix: detect dependency through principal literal

obycode:
The previous pass to detect dependencies between contracts did not
handle the case where another contract is just passed as a contract call
parameter. For example, in the code below, `foo` should be identified as
a dependency.

```clarity
(define-public (pass-foo)
    (ok (print .foo))
)
```


lgalabru (approved):
LGTM, thank you for this improvement @obycode!

PR #118, opened 2022-03-29
feat: add AST-based dependency checker

obycode:
With this enhancement, contracts from a project can be read in in any
order, and after parsing them into ASTs, the dependencies between them
are detected and used to order the contracts for interpretation.


src/repl/interpreter.rs
```diff
@@ -171,15 +171,68 @@ impl ClarityInterpreter {
         Ok(result)
     }
 
+    pub fn run_ast(
```
lgalabru, line 174: Something that would be useful for both the LSP and Orchestra, would be the ability to decouple the analysis from the execution. At the moment, the contract is being saved in the AnalysisDatabase after being executed (which is an unnecessary step that we could avoid in these 2 products), so this would need to change. 
I recently introduced the method
https://github.com/hirosystems/clarity-repl/blob/d3eaac63ee32d0d07aad33ec5ea9315122c9f535/src/repl/interpreter.rs#L342
that could help.

obycode: Thanks for the reminder about that. Is your suggestion that we can just remove this new `run_ast` method, and instead use `save_contract` and `execute` in its place?

obycode:
A future enhancement would perform the ordering on the `requirements`. I will open an issue for that. For now, they should be listed in order in Clarinet.toml.

src/analysis/ast_dependency_detector.rs
```diff
@@ -0,0 +1,974 @@
...
+
+        let mut graph = Graph::new();
+        for (contract, contract_dependencies) in dependencies {
+            let contract_id = lookup.get(contract).unwrap();
+            graph.add_node(*contract_id);
+            for dep in contract_dependencies.iter() {
+                let dep_id = match lookup.get(dep) {
+                    Some(id) => id,
+                    None => {
+                        println!(
```
lgalabru, line 114: When using the repl and clarinet as a library is a desktop app or a server, these `process::exit` are pretty incovenient.
I'm working on getting rid of these pattern across the board, could we return an error here instead?

src/analysis/ast_dependency_detector.rs
```diff
@@ -0,0 +1,974 @@
...
+            for index in deps.iter() {
+                let contract = reverse_lookup[*index];
+                contracts.push(contract.name.as_str());
+            }
+            println!(
+                "{}: cycling dependencies: {}",
+                red!("error"),
+                contracts.join(", ")
+            );
+            process::exit(1);
```
lgalabru, line 143: When using the repl and clarinet as a library is a desktop app or a server, these process::exit are pretty incovenient.
I'm working on getting rid of these pattern across the board, could we return an error here instead?

lgalabru (commented):
It looks great to me, thank you for addressing this @obycode!
For a change as deep as this one, before merging could we run a quick performance benchmark? If we execute the citycoin's or arkadiko's test suite with the current version and this new approach, are we seeing a performance hit? - timing are displayed when executing `clarinet test`.

src/analysis/ast_dependency_detector.rs
lgalabru: Note: if these code paths are not reachable you can use the macro `unreachable!()`

obycode: This case is reachable if there is an unmet dependency, so I think returning a `Result` probably makes sense here.

src/repl/interpreter.rs
lgalabru: Apologies for the latency, let's disregard my comment!

src/analysis/ast_dependency_detector.rs
obycode: Addressed with the last commit.

src/analysis/ast_dependency_detector.rs
obycode: Addressed with the last commit.

lgalabru (approved):
👍

PR #129, opened 2022-04-01
fix(debugger): improve handling of `finish`

obycode:
Instead of stopping at the end of the finished expression, it now stops
at the beginning of the next expression. This makes more sense.


lgalabru (approved):
LGTM!

PR #138, opened 2022-04-29
feat: add reload command

moodmosaic:
Reloads the existing contract(s) in the session, as per https://github.com/hirosystems/clarity-repl/issues/15#issuecomment-1086829948.

### DIFF
--- a/src/repl/session.rs
+++ b/src/repl/session.rs
@@ -696,8 +696,14 @@ impl Session {
 
         for existing in contracts {
             let reloaded = match fs::read_to_string(&existing.path) {
-                Ok(xs) => xs,
-                Err(_) => existing.code,
+                Ok(code) => code,
+                Err(err) => {
+                    output.push(red!(format!(
+                        "Error: unable to read {}: {}",
+                        existing.path, err
+                    )));
+                    existing.code
+                }
             };
 
             self.settings.initial_contracts.push(InitialContract {
--- a/src/repl/session.rs
+++ b/src/repl/session.rs
@@ -698,11 +698,13 @@ impl Session {
             let reloaded = match fs::read_to_string(&existing.path) {
                 Ok(code) => code,
                 Err(err) => {
-                    output.push(red!(format!(
-                        "Error: unable to read {}: {}",
-                        existing.path, err
-                    )));
-                    existing.code
+                    println!(
+                        "{}: Unable to read {}: {}",
+                        red!("error"),
+                        existing.path,
+                        err
+                    );
+                    std::process::exit(1);
                 }
             };
 
--- a/src/repl/session.rs
+++ b/src/repl/session.rs
@@ -579,6 +579,8 @@ impl Session {
             cmd if cmd.starts_with("::decode") => self.decode(&mut output, cmd),
             #[cfg(feature = "cli")]
             cmd if cmd.starts_with("::debug") => self.debug(&mut output, cmd),
+            #[cfg(feature = "cli")]
+            cmd if cmd.starts_with("::reload") => self.reload(&mut output),
 
             snippet => {
                 if self.show_costs {
@@ -677,6 +679,46 @@ impl Session {
             Err(result) => result,
         };
         output.append(&mut result);
+    }
+
+    #[cfg(feature = "cli")]
+    fn reload(&mut self, output: &mut Vec<String>) {
+        self.asts.clear();
+        self.contracts.clear();
+        self.costs_reports.clear();
+        self.coverage_reports.clear();
+        self.executed.clear();
+        self.initial_contracts_analysis.clear();
+        self.interpreter = ClarityInterpreter::new(
+            self.interpreter.get_tx_sender(),
+            self.settings.repl_settings.clone(),
+        );
+        let contracts = self.settings.initial_contracts.clone();
+        self.settings.initial_contracts.clear();
+
+        for existing in contracts {
+            let reloaded = match fs::read_to_string(&existing.path) {
+                Ok(xs) => xs,
+                Err(_) => existing.code,
+            };
+
+            self.settings.initial_contracts.push(InitialContract {
+                code: reloaded,
+                path: existing.path,
+                name: existing.name,
+                deployer: existing.deployer,
+            });
+        }
+
+        match self.start() {
+            Ok(_) => {
+                self.get_contracts(output);
+                output.push("Ok, contracts reloaded.".to_string());
+            }
+            Err(e) => {
+                println!("{}", e);
+            }
+        }
     }
 
     pub fn formatted_interpretation_ast(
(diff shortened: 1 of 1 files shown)


moodmosaic:
This can help pursue a more REPL-focused workflow. The advantage is twofold.

* Use `::reload` as a quick way to type-check syntax of contracts without exiting the console
  image

* Use `::reload` as quick way to get the latest changes in the current loaded contract(s)
  image
  Notice the two occurrences of `::reload` above:
  → `::reload` shows function added in `example` contract :eyes: 
  → `::reload` shows function removed in `key-health-check` contract :eyes:

src/repl/session.rs
```diff
@@ -679,6 +680,47 @@ impl Session {
         output.append(&mut result);
     }
 
+    #[cfg(feature = "cli")]
+    fn reload(&mut self, output: &mut Vec<String>) {
+        self.asts.clear();
+        self.contracts.clear();
+        self.costs_reports.clear();
+        self.coverage_reports.clear();
+        self.executed.clear();
```
obycode, line 689: `executed` is just a list of the commands which have been executed. Let's not clear that one.

src/repl/session.rs
```diff
@@ -679,6 +680,47 @@ impl Session {
...
+        self.initial_contracts_analysis.clear();
+        self.interpreter =
+            ClarityInterpreter::new(
+                self.interpreter.get_tx_sender(),
+                self.settings.repl_settings.clone());
+        let contracts = self.settings.initial_contracts.clone();
+        self.settings.initial_contracts.clear();
```
obycode, line 696: I believe you can avoid this clone by using:
```rust
        let contracts = std::mem::take(&mut self.settings.initial_contracts);
```

src/repl/session.rs
```diff
@@ -679,6 +680,47 @@ impl Session {
...
+
+        for existing in contracts {
+            let reloaded = match fs::read_to_string(&existing.path) {
+                Ok(xs) => xs,
+                Err(_) => existing.code
```
obycode, line 701: I think this should probably report an error.

src/repl/session.rs
```diff
@@ -679,6 +680,47 @@ impl Session {
...
+        self.interpreter =
+            ClarityInterpreter::new(
+                self.interpreter.get_tx_sender(),
+                self.settings.repl_settings.clone());
+        let contracts = self.settings.initial_contracts.clone();
+        self.settings.initial_contracts.clear();
+
+        for existing in contracts {
+            let reloaded = match fs::read_to_string(&existing.path) {
+                Ok(xs) => xs,
```
obycode, line 700: How about a more clear variable name here, like `code`.

obycode (changes requested):
Looking good! Thanks @moodmosaic! Just a few minor changes requested.

src/repl/session.rs
moodmosaic: Thank for the info, I wasn't aware of that. 👍

src/repl/session.rs
moodmosaic: Oh my Haskell 👀 — I apologize for this. Yes, `code` looks better.

src/repl/session.rs
moodmosaic: Very nice. Not only I can replace L965 with `std::mem::take` but I can also get rid of `clear` in L696, according to the [docs](https://doc.rust-lang.org/std/mem/fn.take.html):

```rust
pub fn take<T>(dest: &mut T) -> T 
where
    T: Default, 

// Replaces dest with the default value of T, returning the previous dest value.
```

obycode: Exactly 👍

src/repl/session.rs
moodmosaic: Hmm, perhaps something like this (typing on the fly; not really sure if it'll compile)

```rust
let reloaded = match fs::read_to_string(&existing.path) {
    Ok(code) => code,
    Err(err) => {
        output.push(
          red!(format!(
            "Error: unable to read {:?}: {}", existing.path, err)));
        return existing.code;
    }
```

moodmosaic: This is a side-effect inside a pattern-match, however. Not sure if that's what we really want.

obycode: I'm okay with a side effect here. Maybe also track this so that later when you print "Ok, contracts reloaded", you can instead say something to indicate that there was a (partial) failure. Something like:

```
3 of 4 contracts successfully reloaded.
```

Note: In your code above, you'll want to change `return existing.code;` to just `existing.code`.

moodmosaic: Yep, it seems to work, I'm just not sure if it's idiomatic enough wrt the rest of the code base:

```rust
let reloaded;
match fs::read_to_string(&existing.path) {
    Ok(code) =>
        reloaded = code,
    Err(err) => {
        reloaded = existing.code;
        output.push(
          red!(format!(
            "Error: unable to reload {:?}: {}", existing.path, err)));
    }
}
```

obycode: I think the more idiomatic way would be:

```rust
            let reloaded = match fs::read_to_string(&existing.path) {
                Ok(code) => code,
                Err(err) => {
                    output.push(red!(format!(
                        "Error: unable to read {}: {}",
                        existing.path, err
                    )));
                    existing.code
                }
            };
```

moodmosaic: > Note: In your code above, you'll want to change `return existing.code;` to just `existing.code`.

Hmm... I tried that but I got `match` arms have incompatible types. I pushed https://github.com/hirosystems/clarity-repl/pull/138/commits/272a6f8f17a6b201a0d825bb562287eb679db0b4 which seems to make `rustc` happy-er 🤔

obycode: Make sure you get rid of the `;`. Is that the problem?

moodmosaic: Yes, indeed. Thank you 👍

Force pushed from [272a6f8 ](https://github.com/hirosystems/clarity-repl/commit/272a6f8f17a6b201a0d825bb562287eb679db0b4)to [44f27a2](https://github.com/hirosystems/clarity-repl/commit/44f27a2330bf096f99a2cf8dbdd4f20b9a1a7b05).

moodmosaic: Maybe also track this so that later when you print "Ok, contracts reloaded", you can instead say something to indicate that there was a (partial) failure. Something like:
> 
> ```
> 3 of 4 contracts successfully reloaded.
> ```

That's a good suggestion. I will look into it. 👍

obycode: It's up to you. I would also be okay with an error reading a contract file being a catastrophic error that exits the console completely, since that likely means that the Clarinet.toml has changed, which requires re-starting anyway.

moodmosaic: Indeed, that sounds better (and it'll be less lines of code also).

moodmosaic: Done now in https://github.com/hirosystems/clarity-repl/pull/138/commits/b76aa37ce5b041bad9d779c07e5733a1463628c7.

And if/when we add support for picking up changes in Clarinet.toml (without re-starting) we can probably revisit this part so that instead of exiting it'll keep track of the contracts that could actually be reloaded.

obycode (dismissed):
LGTM. Thanks @moodmosaic!!

obycode:
@moodmosaic please fix the formatting which caused the CI failure, then this should be good to merge. See [rustfmt](https://github.com/rust-lang/rustfmt).

moodmosaic:
@obycode, done. `cargo fmt --all -- --check` should yield no messages now. Rebased and force-pushed.

obycode (approved):

obycode (approved):

PR #142, opened 2022-05-02
Add DAP debugger

obycode:


obycode:
Oops, forgot to update the dependency on dap-types. Fixing now...

obycode:
> Oops, forgot to update the dependency on dap-types. Fixing now...

Fixed.

src/clarity/mod.rs
```diff
@@ -82,6 +83,31 @@ impl_array_newtype!(BurnchainHeaderHash, u8, 32);
 impl_array_hexstring_fmt!(BurnchainHeaderHash);
 // impl_byte_array_newtype!(BurnchainHeaderHash, u8, 32);
 
+/** EvalHook defines an interface for hooks to execute during evaluation. */
+pub trait EvalHook {
+    // Called before the expression is evaluated
+    fn begin_eval(
```
lgalabru, line 89: I wonder if `will_begin_eval` would be more explicit, what do you think?

src/clarity/mod.rs
```diff
@@ -82,6 +83,31 @@ impl_array_newtype!(BurnchainHeaderHash, u8, 32);
...
+        &mut self,
+        env: &mut Environment,
+        context: &LocalContext,
+        expr: &SymbolicExpression,
+    ) {
+    }
+
+    // Called after the expression is evaluated
+    fn finish_eval(
```
lgalabru, line 98: I wonder if `didl_finish_eval` would be more explicit, what do you think?

src/clarity/mod.rs
```diff
@@ -82,6 +83,31 @@ impl_array_newtype!(BurnchainHeaderHash, u8, 32);
...
+        &mut self,
+        env: &mut Environment,
+        context: &LocalContext,
+        expr: &SymbolicExpression,
+        res: &core::result::Result<Value, crate::clarity::errors::Error>,
+    ) {
+    }
+
+    // Called upon completion of the execution
+    fn complete(&mut self, result: core::result::Result<&mut ExecutionResult, String>) {}
```
lgalabru, line 108: What is the difference between `finish_eval` and `complete`? (I will probably figure it out later in the review)

obycode: Yes, you will see, but `complete` runs when the entire execution is finished. Unfortunately, I wasn't able to make it work without this additional hook, but it might be possible. Let me know if you see a better solution.

src/clarity/debug/dap/mod.rs
```diff
@@ -0,0 +1,1146 @@
...
+            supports_configuration_done_request: Some(true),
+            supports_function_breakpoints: Some(true),
+            supports_step_in_targets_request: Some(true),
+            support_terminate_debuggee: Some(true),
+            supports_loaded_sources_request: Some(true),
+            supports_data_breakpoints: Some(true),
+            supports_breakpoint_locations_request: Some(true),
+            supports_conditional_breakpoints: None,
+            supports_hit_conditional_breakpoints: None,
+            supports_evaluate_for_hovers: None,
```
lgalabru, line 264: this capability would be pretty cool!

src/clarity/debug/dap/mod.rs
```diff
@@ -0,0 +1,1146 @@
...
+            request_seq: seq,
+            success: true,
+            message: None,
+            body: Some(ResponseBody::Launch),
+        });
+
+        false
+    }
+
+    fn setBreakpoints(&mut self, seq: i64, arguments: SetBreakpointsArguments) -> bool {
```
lgalabru, line 402:
```suggestion
    fn setBreakpoints(&mut self, seq: i64, arguments: SetBreakpointsArguments) -> bool {
    fn set_breakpoints(&mut self, seq: i64, arguments: SetBreakpointsArguments) -> bool {
```

src/clarity/debug/dap/mod.rs
```diff
@@ -0,0 +1,1146 @@
...
+
+        self.send_response(response);
+        false
+    }
+
+    fn quit(&mut self, seq: i64, arguments: DisconnectArguments) -> bool {
+        // match arguments.restart {
+        //     Some(restart) => restart,
+        //     None => false,
+        // }
```
lgalabru, line 794: Dead code?

src/clarity/debug/dap/mod.rs
```diff
@@ -0,0 +1,1146 @@
...
+                    Some(path) => path.clone(),
+                    _ => "debugger".to_string(),
+                },
+            ),
+            source_reference: None,
+            presentation_hint: None,
+            origin: None,
+            sources: None,
+            adapter_data: None,
+            checksums: None,
```
lgalabru, line 975: Speaking of checksums, I was actually wondering about this, how is the DAP protocol dealing with "stale" data? Let say I add a breakpoint using VSCode UI, delete the line, are we updating the state automatically?

lgalabru (commented):
🏆 This looks great, thank you @obycode. I'm so excited by this feature! I had a couple of non blocking questions after a first pass, I will give it a second look.

src/clarity/debug/dap/mod.rs
obycode: Agreed. And it's probably not difficult to add. I'll open an issue to do it after the initial release.

obycode: Actually, I remembered this seems to already work, even though it is not specified here. But maybe this flag enables something more.

image

src/clarity/debug/dap/mod.rs
obycode: Oops 😅. My brain is not fully in Rust mode sometimes.

src/clarity/debug/dap/mod.rs
obycode: yes. I'll add an issue to support restart later too.

src/clarity/debug/dap/mod.rs
obycode: Good question. No, currently, if a user changes the source in the middle of a debug session, the original code is still debugged. That would be a cool feature to support later. Opening another issue to remember that 😄

lgalabru (approved):
LGTM, thanks @obycode!

PR #143, opened 2022-05-03
fix: harden `ASTVisitor` to handle invalid ASTs

obycode:
Originally, `ASTVisitor` was only used after the type checker, so it could
be assumed that the AST was well-formed. Now, we are using it very
early, to detect dependencies in the AST. This changeset avoids crashing
by replacing all prior uses of `unwrap()` with `unwrap_or` or
`unwrap_default`, so that a placeholder is used in the case where the
expected type is not present.

Fixes: hirosystems/clarinet#334

### DIFF
--- a/src/analysis/ast_visitor.rs
+++ b/src/analysis/ast_visitor.rs
@@ -13,6 +13,13 @@ pub struct TypedVar<'a> {
     pub decl_span: Span,
 }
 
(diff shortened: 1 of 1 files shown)


lgalabru (approved):
LGTM, thanks @obycode!

Issue #145, opened 2022-05-04
Runtime error when calling contract with too many dots


friedger:

obycode:
Thanks for reporting this @friedger. What's happening is that the v2 parser is treating `SP3K8BC0PPEVCV7NZ6QSRWPQ2JE9E5B6N3PA0KBR9.SP3K8BC0PPEVCV7NZ6QSRWPQ2JE9E5B6N3PA0KBR9.fixed-weight-pool-v1-01` as a trait identifier (principal.contract.name), but it did not check the 40 character contract name limit. The original parser doesn't match this as a trait identifier due to this problem so it fails. I'll add the character limit check to the v2 parser.

I think I had assumed that the `ContractName::try_from` would have failed if it was not a valid contract name. The regex in src/clarity/ast/parser/mod.rs has this size limit but the regex in src/clarity/representation.rs doesn't. I'll add it there so that we get the same behavior.

src/clarity/representation.rs:
```rust
    pub static ref CONTRACT_NAME_REGEX: Regex =
        Regex::new("^[a-zA-Z]([a-zA-Z0-9]|[-_])*$|^__transient$").unwrap();
```

src/clarity/ast/parser/mod.rs:
```
pub const CONTRACT_MIN_NAME_LENGTH: usize = 1;
pub const CONTRACT_MAX_NAME_LENGTH: usize = 40;
...
    pub static ref CONTRACT_NAME_REGEX: String = format!(
        r#"([a-zA-Z](([a-zA-Z0-9]|[-_])){{{},{}}})"#,
        CONTRACT_MIN_NAME_LENGTH - 1,
        CONTRACT_MAX_NAME_LENGTH - 1
    );
```

PR #146, opened 2022-05-04
fix: check the maximum length of a contract name

obycode:
The parser used `ClarityName::try_from` to check for a valid name, but
this check did not check the length of the name (must be 1-40
characters).

Fixes: #145


lgalabru (commented):
Hmm I'm ambivalent about this fix. You're tweaking a consensus breaking regex, and the construction of `ContractName`. The scope of this impact and the eventual divergence with a canonical behavior is unclear to me. Is there another way to fix this issue?

obycode:
I don't think this is actually consensus breaking. The canonical parser already applies this 40-character limit in a different regex:

src/clarity/ast/parser/mod.rs:
```
pub const CONTRACT_MIN_NAME_LENGTH: usize = 1;
pub const CONTRACT_MAX_NAME_LENGTH: usize = 40;
...
    pub static ref CONTRACT_NAME_REGEX: String = format!(
        r#"([a-zA-Z](([a-zA-Z0-9]|[-_])){{{},{}}})"#,
        CONTRACT_MIN_NAME_LENGTH - 1,
        CONTRACT_MAX_NAME_LENGTH - 1
    );
```

This regex in representations.rs is only used in `ContractName::try_from`, which is used in places where the string is already known to be a valid contract name, and in the codec for `consensus_deserialize`. 

What if instead of this change I use the existing parser regex?

```
    pub static ref CONTRACT_NAME_REGEX: Regex =
        Regex::new(&format!("^{}$|^__transient$", *parser::CONTRACT_NAME_REGEX)).unwrap();
```

obycode:
On second thought, the new parser does not require this regex, so then I would prefer to leave the regex in this PR, perhaps with the modification to use constants for the length limits. What do you think @lgalabru?

PR #147, opened 2022-05-05
feat: better simulated block time

obycode:
The original implementation always returns 0 for block time, but some
developers require a slightly more realistic number. This commit just
returns the block height multiplied by 600 (10 minutes) to provide a
monotonic increasing value for this time.

Fixes: hirosystems/clarinet#310


lgalabru (approved):
LGTM, thanks @obycode!

luizalbertocviana:
Thank you guys.
That will unblock several tasks on my side :)

PR #151, opened 2022-05-10
feat: return unresolved dependencies from detector

obycode:


src/analysis/ast_dependency_detector.rs
```diff
@@ -46,14 +48,105 @@ pub struct ASTDependencyDetector<'a> {
...
 
+#[derive(Clone, Debug, Eq)]
+pub struct Dependency {
+    pub contract_id: QualifiedContractIdentifier,
+    pub required_before_publish: bool,
+}
+
+impl PartialEq for Dependency {
+    fn eq(&self, other: &Self) -> bool {
+        self.contract_id == other.contract_id
```
lgalabru, line 63: Shouldn't we including `required_before_publish` in this equality?

obycode: See the `add_dependency` method. What I was going for is that a contract is only in the set once (regardless of this flag), and a dependency with the `required_before_publish` flag overwrites a dependency without that flag. Is this a confusing way to do that?

lgalabru: Works for me, thank you unpacking!

lgalabru (approved):
LGTM!

PR #159, opened 2022-05-18
Removing requirements / links features from the REPL

lgalabru:
With this PR, we are getting rid of the requirements, we came to the conclusion that a more natural place would be clarinet.
There a few other adjustments / additions / fixes that were made along the way of working on https://github.com/hirosystems/clarinet/pull/333, that are required and expected by this PR.

### DIFF
--- a/src/clarity/diagnostic.rs
+++ b/src/clarity/diagnostic.rs
@@ -3,7 +3,7 @@ use std::fmt;
 
 /// In a near future, we can go further in our static analysis and provide different levels
 /// of diagnostics, such as warnings, hints, best practices, etc.
-#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub enum Level {
     Note,
     Warning,
@@ -28,7 +28,7 @@ pub trait DiagnosableError {
     }
 }
 
-#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct Diagnostic {
     pub level: Level,
     pub message: String,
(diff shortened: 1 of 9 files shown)


src/clarity/diagnostic.rs
```diff
@@ -71,55 +71,55 @@ impl Diagnostic {
...
             let first_line = span.start_line.saturating_sub(1) as usize;
             let last_line = span.end_line.saturating_sub(1) as usize;
 
             output.push(lines[first_line].clone());
-            let mut pointer = format!(
-                "{: <1$}^",
-                "",
-                (span.start_column.saturating_sub(1)) as usize
-            );
+            let mut pointer = format!("{: <1$}^", "", (span.start_column - 1) as usize);
```
obycode, line 109: What is the reason for this change from `saturating_sub(1)` to `- 1`. I think it was using the saturating sub so that it correctly handles the case where the start_column is 0.

src/repl/session.rs
```diff
@@ -536,6 +343,47 @@ impl Session {
...
+        );
+
+        let bns_contract = ContractName::try_from("bns").unwrap();
+        let pox_contract = ContractName::try_from("pox").unwrap();
+
+        let boot_testnet =
+            PrincipalData::parse_standard_principal("ST000000000000000000002AMW42H").unwrap();
+
+        let boot_mainnet =
+            PrincipalData::parse_standard_principal("SP000000000000000000002Q6VF78").unwrap();
```
obycode, line 366: Would it make sense to move these into lazy statics?

obycode (approved):
Looks good to me, just two questions.

src/clarity/diagnostic.rs
lgalabru: Addressed with [latest](https://github.com/hirosystems/clarity-repl/pull/159/commits/257b12af8673c56d8112232e2489b564beaf5755)

src/repl/session.rs
lgalabru: Addressed with [latest](https://github.com/hirosystems/clarity-repl/pull/159/commits/257b12af8673c56d8112232e2489b564beaf5755)

obycode (approved):
LGTM!

obycode (approved):

PR #160, opened 2022-05-18
fix: make dep detector robust to invalid ASTs

obycode:
Since the `ASTDependencyDetector` is now used before any checks have
been performed on the AST, it cannot make assumptions about correctness
and crash when that assumption is wrong. For example, when unwrapping an
expected type, we also provide a default value in case the type was
incorrect in the AST.


lgalabru (approved):
LGTM, thank you @obycode!

PR #180, opened 2022-05-26
fix: ability to interpret by previous ASTs

lgalabru:


lgalabru:
I think we have a lot of cleaning to perform in the session / interpreter, I can see a lot of dead code path.

PR #185, opened 2022-06-02
Read expressions from a file

fariedt:
Adds the `::read` command to the console.

### DIFF
--- a/src/repl/session.rs
+++ b/src/repl/session.rs
@@ -452,6 +452,7 @@ impl Session {
         output
     }
 
+    #[cfg(feature = "cli")]
     fn run_snippet(&mut self, output: &mut Vec<String>, cost_track: bool, cmd: &str) {
         let (mut result, cost) = match self.formatted_interpretation(
             cmd.to_string(),
@@ -1129,10 +1130,9 @@ impl Session {
         output.push(green!(format!("{}", value)));
     }
 
-    #[cfg(feature = "cli")]
     pub fn get_costs(&mut self, output: &mut Vec<String>, cmd: &str) {
         let snippet = cmd.to_string().split_off("::get_costs ".len());
-        self.run_snippet(output, self.show_costs, &snippet.to_string());
+        self.run_snippet(output, true, &snippet.to_string());
     }
 
     #[cfg(feature = "cli")]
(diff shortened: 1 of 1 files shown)


src/repl/session.rs
```diff
@@ -625,6 +627,26 @@ impl Session {
...
+
+        let snippet = match fs::read_to_string(filename) {
+            Ok(snippet) => snippet,
+            Err(err) => return output.push(red!(format!("unable to read {}: {}", filename, err))),
+        };
+
+        let mut result = match self.formatted_interpretation(snippet, None, false, None, None) {
+            Ok((output, _)) => output,
+            Err(result) => result,
+        };
```
obycode, line 645: After reading from the file, I think it should be treated the same as when a snippet is passed in. The missing piece here then would be handling when `show_costs` mode is enabled.

Perhaps a clean solution would be to refactor the `snippet => {...}` case of the match in `handle_command` into a function, then call that function here.

obycode (changes requested):
Thanks for adding this @fariedt! I just have one change request.

src/repl/session.rs
fariedt: I've pushed a commit for this.  It's a clean change; I was able to merge both `handle_command` and the cli `get_costs` but not the non-cli `get_costs`.  The main change in behavior from my first commit is that `::read` can now show the execution costs if the user has toggled cost reporting.

obycode (changes requested):
Thanks for this change. I approved, but now I see that the wasm build is failing because of that change. I guess we do need two versions of `run_snippet` for the cli and non-cli modes. That should then allow you to get merge `get_costs` completely into these new methods.

src/repl/session.rs
obycode: See new review comment.

src/repl/session.rs
```diff
@@ -1202,6 +1217,12 @@ impl Session {
         output.append(&mut result);
     }
 
+    #[cfg(not(feature = "cli"))]
+    pub fn get_costs(&mut self, output: &mut Vec<String>, cmd: &str) {
+        let snippet = cmd.to_string().split_off("::get_costs ".len());
+        self.run_snippet(output, true, &snippet.to_string());
+    }
```
obycode, line 1224: Looks like the two versions of `get_costs` are identical now, so we can merge into one and remove the feature gating.

obycode (changes requested):

obycode (approved):
Looks great. Thanks!

PR #188, opened 2022-06-03
fix: remove unhelpful dependency error

obycode:
Instead of reporting an error in `order_contracts` when a contract is
not found, we now just skip that dependency. Reporting an error at that
point is not useful, because a better error with proper location
information can be reported by the later analysis.

See hirosystems/clarinet#396


lgalabru (approved):
LGTM!

PR #190, opened 2022-06-09
fix: don't create a dependency for `__transient`

obycode:
The `__transient` contract is used as a placeholder when the principal
is invalid, so just ignore this in the dependency detector, and a proper
error will be generated for it.


lgalabru (approved):
Thanks @obycode!

PR #192, opened 2022-06-21
fix: avoid crash from too few args passed

obycode:
If an argument is missing in a call, and that parameter has a trait
type, the dependency checker would crash when trying to retrieve the
argument to check it. This fix checks the argument count first to skip
this case, since the error should be reported by the call-checker.


lgalabru (approved):
Thanks @obycode!

PR #264, opened 2023-05-24
fix: remove undeclared dependency

utibeabasi6:


utibeabasi6:
Hopefully this should fix the ci pipeline and publish the docker image

lgalabru:
hey @utibeabasi6! if we want to publish a docker image, we'll have to implement that action :)
we will address this soon!

PR #266, opened 2023-05-25
feat: attempt to support cursed inscriptions

lgalabru:
We are test driving the concept of [cursed inscriptions](https://github.com/casey/ord/issues/2045). This will probably take multiple iteration. 
With this PR, our goal is to create cursed inscription for transactions that tried to inscribe satoshis, using other inputs than input #0.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3403,9 +3403,9 @@ dependencies = [
 
 [[package]]
 name = "rocket_okapi"
-version = "0.8.0-rc.2"
+version = "0.8.0-rc.1"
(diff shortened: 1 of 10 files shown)


lgalabru:
Gave a first shot, ran into the following issue and ran into at least one issue.
Will follow up based on the outcome of https://github.com/casey/ord/issues/2045#issuecomment-1565452803.

lgalabru:
Looks like the approach I suggested was also implemented in `ord`.
Green data env is looking good, merging!

PR #297, opened 2023-06-27
feat: add metrics to `/ping` response of event observer server

### LINKED ISSUE #285: Augment /ping endpoint with health related metrics
Schema prototype:

```json
{
  "chain_tips": {
      "stacks": <BlockIdentifier>,
      "bitcoin": <BlockIdentifier>,
  },
  "running_predicates": <integer>,
  "last_block_ingestion_at": <DateTime>
}
```

MicaiahReid:
Fixes #285

### DIFF
--- a/components/chainhook-sdk/src/observer/mod.rs
+++ b/components/chainhook-sdk/src/observer/mod.rs
@@ -762,9 +762,9 @@ pub async fn start_observer_commands_handler(
                             .as_millis()
                             .into();
                     }
-                    Err(e) => {
-                        ctx.try_log(|logger| slog::warn!(logger, "Failed to aquire lock: {}", e))
-                    }
+                    Err(e) => ctx.try_log(|logger| {
+                        slog::warn!(logger, "unable to acquire observer_metrics_rw_lock:{}", e)
+                    }),
                 };
                 bitcoin_block_store.insert(new_block.block_identifier.clone(), new_block);
             }
@@ -1022,7 +1022,11 @@ pub async fn start_observer_commands_handler(
                                         highest_tip_block.timestamp.into();
                                 }
                                 Err(e) => ctx.try_log(|logger| {
-                                    slog::warn!(logger, "Failed to aquire lock: {}", e)
+                                    slog::warn!(
+                                        logger,
+                                        "unable to acquire observer_metrics_rw_lock:{}",
+                                        e
+                                    )
                                 }),
                             },
                             None => {}
@@ -1166,7 +1170,11 @@ pub async fn start_observer_commands_handler(
                         match observer_metrics.write() {
                             Ok(mut metrics) => metrics.bitcoin.deregister_prediate(),
                             Err(e) => ctx.try_log(|logger| {
-                                slog::warn!(logger, "Failed to aquire lock: {}", e)
+                                slog::warn!(
+                                    logger,
+                                    "unable to acquire observer_metrics_rw_lock:{}",
+                                    e
+                                )
                             }),
                         }
                     }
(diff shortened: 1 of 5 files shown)


components/chainhook-sdk/src/observer/mod.rs
```diff
@@ -418,6 +433,9 @@ pub async fn start_event_observer(
 
     let background_job_tx_mutex = Arc::new(Mutex::new(observer_commands_tx.clone()));
 
+    let observer_metrics = ObserverMetrics::default();
+    let observer_metrics_arc = Arc::new(Mutex::new(observer_metrics));
```
lgalabru, line 437: Could we use a `RwLock`, instead of a `Mutex`?
The API is always reading, and the runloop writing.

lgalabru (commented):
Thank you @MicaiahReid! Just a quick adjustment and we should be good to go!

MicaiahReid:
Thanks @lgalabru, I was wondering about that. I've made that update! I've still got some more stuff I want to add to this, I'll mark it as ready to review when it's done!

MicaiahReid:
@lgalabru I think it's ready for a review!

I've tested everything except the reorg metrics. If you have thoughts on how to test that I'd love to!

components/chainhook-sdk/src/observer/mod.rs
```diff
@@ -363,6 +363,35 @@ impl ChainhookStore {
...
+    pub fn deregister_prediate(&mut self) {
+        self.registered_predicates -= 1;
+        self.deregistered_predicates += 1;
+    }
+}
+
+#[derive(Debug, Default, Serialize, Clone)]
+pub struct ObserverMetrics {
+    pub bitcoin: ChainMetrics,
+    pub stacks: ChainMetrics, // todo: do users sometimes start with just bitcoin, making stacks unnecessary?
```
MicaiahReid, line 392: @lgalabru

I can remove the todo here, I just wanted to confirm that we're okay with `stacks` being a permanent member of these metrics. It looks like for most of this file we're assuming that stacks is enabled, but maybe I'm missing something and should make the stacks metrics optional.

components/chainhook-sdk/src/observer/mod.rs
```diff
@@ -1142,6 +1236,67 @@ pub async fn start_observer_commands_handler(
...
+                                        "unable to acquire observer_metrics_rw_lock:{}",
+                                        e
+                                    )
+                                }),
+                            },
+                            None => {}
+                        }
+                    }
+                    StacksChainEvent::ChainUpdatedWithReorg(update) => {
+                        // todo: I don't know which block's timestamp we actually want
```
MicaiahReid, line 1273: @lgalabru 
Another thing I wanted your confirmation on: for reorg time, do we want the timestamp of the highest-tipped new block we're applying, or do we just want `time.now()`

components/chainhook-sdk/src/observer/mod.rs
MicaiahReid: Hmm, I wondering if other stuff would have to change for that to make sense. It seems the event observer always assumes a stacks chain is running. All of the configuration has stacks data as non-optional, and we're always exposing stacks routes:
```
let mut routes = rocket::routes![
        http::handle_ping,
        http::handle_new_bitcoin_block,
        http::handle_new_stacks_block,
        http::handle_new_microblocks,
        http::handle_new_mempool_tx,
        http::handle_drop_mempool_tx,
        http::handle_new_attachement,
        http::handle_mined_block,
        http::handle_mined_microblock,
    ];
```

Is that a correct assessment? If so, should I create an issue to remove logic from the event observer that always enables stacks event observation?

lgalabru: My apologies for the confusion. If the event observer is started, then we know for sure that Stacks + Bitcoin are observed - we don't (or should not) start the server if the bitcoin block signaling is happening through Zeromq.

MicaiahReid: Okay great, thanks. I've removed the TODO

lgalabru (approved):
Amazing, thank you for adding some tests!

PR #299, opened 2023-06-28
chore: add zmq url to default chainhook config

MicaiahReid:


components/chainhook-cli/src/config/generator.rs
```diff
@@ -16,7 +16,13 @@ mode = "mainnet"
 bitcoind_rpc_url = "http://localhost:8332"
 bitcoind_rpc_username = "devnet"
 bitcoind_rpc_password = "devnet"
+# Bitcion block events can be received by Chainhook
```
lgalabru, line 19:
```suggestion
# Bitcoin block events can be received by Chainhook
```

lgalabru (approved):
Thanks @MicaiahReid 🙌

PR #302, opened 2023-06-29
feat: upgrade typescript client

rafaelcr:
* Send HTTP 500 response on payload errors so node can retry
* Fix `curse_type` schema to include null and tagged objects
* Add `wait_for_chainhook_node` option to be able to skip server pings on local chainhook tests


lgalabru (approved):
LGTM!

PR #307, opened 2023-06-30
fix: transfer refactor

lgalabru:
In this PR, we're revisiting (drastically) the way ordinals transfers are being tracked.
Instead of only keeping the location of the last outpoint, we're creating a dedicated entry in a new `locations` sqlite table.
We could be pruning the locations at some point for optimizing storage, but the current focus has to be on correctness, vs optimization.


components/chainhook-cli/src/service/mod.rs
```diff
@@ -116,22 +118,71 @@ impl Service {
...
+                            for bitcoin_predicate in chainhook_config.bitcoin_chainhooks.iter_mut()
+                            {
+                                bitcoin_predicate.enabled = false;
+                            }
+                            let mut bitcoin_predicates_ref: Vec<
+                                &chainhook_sdk::chainhooks::types::BitcoinChainhookSpecification,
+                            > = vec![];
+                            for bitcoin_predicate in chainhook_config.bitcoin_chainhooks.iter() {
+                                bitcoin_predicates_ref.push(bitcoin_predicate);
+                            }
```
MicaiahReid, line 158: Would it be possible for us just to loop once here?

```suggestion
                            let mut bitcoin_predicates_ref: Vec<
                                &chainhook_sdk::chainhooks::types::BitcoinChainhookSpecification,
                            > = vec![];
                            for bitcoin_predicate in chainhook_config.bitcoin_chainhooks.iter_mut()
                            {
                                bitcoin_predicate.enabled = false;
                                bitcoin_predicates_ref.push(bitcoin_predicate);
                            }
```

components/chainhook-sdk/src/hord/tests/mod.rs
```diff
@@ -0,0 +1 @@
+
```
MicaiahReid, line 1: 🥇

MicaiahReid (changes requested):

components/chainhook-cli/src/config/mod.rs
```diff
@@ -401,13 +406,13 @@ impl Config {
...
             },
             network: IndexerConfig {
-                stacks_node_rpc_url: "http://0.0.0.0:20443".into(),
                 bitcoind_rpc_url: "http://0.0.0.0:18443".into(),
                 bitcoind_rpc_username: "devnet".into(),
                 bitcoind_rpc_password: "devnet".into(),
-                bitcoin_block_signaling: BitcoinBlockSignaling::Stacks(
-                    "http://0.0.0.0:20443".into(),
-                ),
+                bitcoin_block_signaling: BitcoinBlockSignaling::Stacks(StacksNodeConfig {
```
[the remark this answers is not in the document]
lgalabru: Addressed with latest

components/chainhook-cli/src/service/mod.rs
lgalabru: Addressed in https://github.com/hirosystems/chainhook/pull/307/commits/ba662e5149ecc9663e50c2f4cf0d1ba49e5486ee

MicaiahReid (approved):

PR #309, opened 2023-07-03
feat: update inscription transfer schemas

rafaelcr:
* Remove `inscription_number` and `ordinal_number` from inscription transfers
* Add option to validate chainhook payload schemas
* Add option to set a custom body size limit


lgalabru (approved):
Thanks @rafaelcr!

PR #321, opened 2023-07-06
fix: stx_event predicate schema

rafaelcr:


components/client/typescript/src/schemas/stacks/if_this.ts
```diff
@@ -31,8 +31,9 @@ export type StacksIfThisNftEvent = Static<typeof StacksIfThisNftEventSchema>;
 
 export const StacksIfThisStxEventSchema = Type.Object({
   scope: Type.Literal('stx_event'),
-  asset_identifier: Type.String(),
-  actions: Type.Array(Type.String()),
+  actions: Type.Array(
+    Type.Union([Type.Literal('mint'), Type.Literal('transfer'), Type.Literal('lock')])
```
BLuEScioN, line 35: Is burn possible?

rafaelcr: Not here but added to nft_event and ft_event

PR #338, opened 2023-07-14
Changed links to use relative links

LakshmiLavanyaKasturi:
Changed absolute links to relative links to fix the build issues.


lgalabru (approved):
Thanks @LakshmiLavanyaKasturi!

PR #331, opened 2023-07-12
Release/v0.17.0

lgalabru:
Addressing a few DX flow  and performance issues + several adjustments following the hord split, along with adding a testing pattern proposal for securing predicates evaluation.

PR #356, opened 2023-07-27
feat: match on all `print_event`s if `contract_identifier` is `None`

### LINKED ISSUE #294: make the `contract_identifier` arg be optional in `print_event` predicate
The `print_event` predicate currently requires the `contract_identifier` argument:

```json5
// Get any transaction emitting given print events predicate
// `contract-identifier` mandatory argument admits:
//  - string type, fully qualifying the contract to observe. example: `ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.monkey-sip09`
// `contains` mandatory argument admits:
//  - string type, used for matching event
{
    "if_this": {
        "scope": "print_event",
        "contract_identifier": "ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.monkey-sip09",
        "contains": "vault"
    },
}
```

However, in order for the Token Metadata API to detect SIP-019 notifications and SFT mint events, we must be able to inspect *all* print events from *all* contracts to see if they conform to the required structure.

Could we make the `contract_identifier` arg be optional so we can monitor all print events on the stacks chain with this predicate if it isn't specified?

MicaiahReid:
This PR fixes #294 by making the `contract_identifier` and `contains` fields on the `print_event` stacks predicate optional. If the `contract_identifier` value is not provided, all `print_event` values are checked for a match. If the `contains` value is not provided, all event values are considered a match.


lgalabru (commented):
Thanks @MicaiahReid!
- We added similar wildcard capabilities in the past for certain predicates (contract deployments I believe), but instead of going with an optional, we went with `*`. I think I personally like the idea of matching against an optional in the code, while keeping the `*` in the if this / then that, but it's not a strong opinion. So no preference for one or the other, but I'm in favor of consistency.  What do you think?
- Following up on the recent series of issues spotted by Tim (I'm fully taking the blame here), I'd be in favor of raising the bar on unit testing.
Could we please add a test for this new case, by building up on the unit testing proposal here: https://github.com/hirosystems/chainhook/pull/331/commits/b8738e10fe1267db08deeb5f68b6d94b0afa31db ?

lgalabru (approved):
Looking great, thanks @MicaiahReid!

PR #359, opened 2023-07-29
fix: coinbase tx sats_in < sats_out

bingryan:
### Description

In my local regtest test environment, I found that condition: `sats_in < sats_out`. The reason is that current block is only coinbase transaction , it have no other inputs, so sats_in is 0 and sats_out > 0

### Checklist

- [x] All tests pass


lgalabru:
Thank you for your time and this contribution @bingryan, this issue was fixed in a stealth PR that just merged!

PR #362, opened 2023-08-02
chore: additional ordhook adjustments 

lgalabru:
In this PR, we're addressing a few improvements bottleneck detected during hord instrumentation / profiling sessions, along with a few other adjustments in the sdk, required for bringing an eventual, incoming, stable `ordhook` release to the finish line.


MicaiahReid (approved):
Nice! Do we want to make an issue to implement the same exponential backoff to the other `..._with_retry` functions?

PR #378, opened 2023-08-10
feat: include `contract_abi` in stacks transaction

MicaiahReid:
### Example

---

### Checklist

- [x] All tests pass


components/chainhook-types-rs/Cargo.toml
```diff
@@ -14,3 +14,4 @@ serde_derive = "1"
 strum = { version = "0.23.0", features = ["derive"] }
 schemars = { version = "0.8.10", git = "https://github.com/hirosystems/schemars.git", branch = "feat-chainhook-fixes" }
 hex = "0.4.3"
+clarity-vm = "2.1.1"
```
MicaiahReid, line 17: @lgalabru I was surprised at how few dependencies we have in `chainhook-types-rs`. Am I safe to use this dependency here, particularly in `rosetta.rs`?

lgalabru: ouch. can we avoid this dependency on `clarity-vm` somehow?

lgalabru: (we could end up with annoying dependency cycles if we're pulling this lib here)

lgalabru (commented):
Thank you for adding these tests @MicaiahReid!
Is there a way we can get avoid pulling `clarity-vm` into `chainhook-types-rs`?
I think I'd be fine with duplicating the code.

PR #380, opened 2023-08-11
feat: allow matching with regex for stacks print_event

### LINKED ISSUE #348: Allow regex for `contains` based chainhooks
When creating a chainhook that listens for a print_event, it'd be great to be able to use regex to define the condition for `contains`. 

Example: I'd like to create a print_event predicate that only listens for `stack-extend` or `stack-increase` but exclude `stack-aggregation-*`

MicaiahReid:
### Description

This PR introduces the ability to match on a `print_event`'s values with a regex string rather than just with the previous `contains` field.

Fixes #348 

### Example

This allows the following new `if_this` condition for a stacks predicate:
```JSON
{
  "scope": "print_event",
  "contract_identifier": "ST3AXH4EBHD63FCFPTZ8GR29TNTVWDYPGY0KDY5E5.loan-data",
  "regex": "... some regex to match on the event data"
}
```

---

### Checklist

- [x] All tests pass

### DIFF
--- a/components/chainhook-sdk/src/chainhooks/types.rs
+++ b/components/chainhook-sdk/src/chainhooks/types.rs
@@ -776,9 +776,17 @@ pub enum StacksTrait {
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, JsonSchema)]
 #[serde(rename_all = "snake_case")]
-pub struct StacksPrintEventBasedPredicate {
-    pub contract_identifier: String,
-    pub contains: String,
+#[serde(untagged)]
+pub enum StacksPrintEventBasedPredicate {
+    Contains {
+        contract_identifier: String,
+        contains: String,
+    },
+    MatchesRegex {
+        contract_identifier: String,
+        #[serde(rename = "matches_regex")]
+        regex: String,
+    },
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, JsonSchema)]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,9 +30,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.20"
+version = "1.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc936419f96fa211c1b9166887b38e5e40b19958e5b895be7c1f93adec7071ac"
+checksum = "86b8f9420f797f2d9e935edf629310eb938a0d839f984e25327f3c7eed22300c"
 dependencies = [
  "memchr",
 ]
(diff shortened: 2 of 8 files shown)


components/chainhook-sdk/src/chainhooks/types.rs
```diff
@@ -776,9 +776,17 @@ pub enum StacksTrait {
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, JsonSchema)]
 #[serde(rename_all = "snake_case")]
-pub struct StacksPrintEventBasedPredicate {
-    pub contract_identifier: String,
-    pub contains: String,
+#[serde(untagged)]
+pub enum StacksPrintEventBasedPredicate {
```
MicaiahReid, line 780: I _**really**_ don't like that we have to duplicate the definition of `contract_identifier` here. If anyone knows a way around this, that would be sick. That's what lead me down the path of the breaking change PR (#379), but I guess it's probably worth it to not break the API so soon after releasing v1.

aulneau:
awesome, this looks great. will make use of it as soon as it can be released/deployed :~)

components/chainhook-sdk/src/chainhooks/types.rs
lgalabru: Agreed. After toying with serde, it does not look like something supported. If we really wanted to avoid the repetition, we could move to something like this:
```rust
struct Predicate {
   contract_identifier: T,
   contains: Option<U>,
   matches: Option<V>
}
```
and associate a method to that struct to make it more usable.

lgalabru (approved):
Looking great, thanks Micaiah!
With this addition, we should probably start thinking about a way to measure predicate evaluation performance.

components/chainhook-sdk/src/chainhooks/types.rs
MicaiahReid: Yeah that is an option.

In this case I guess we could treat this as an "OR", so we match any that contain _or_ regex match if both values are provided. I think it maybe sets a weird precedence for how we do this in the future - it would feel clunky to have a predicate type with many Options that are only really intended to be used alone, but the user can specify all of them. And if we ever need to make them mutually exclusive (you can have `fieldA` OR `fieldB` specified in the predicate, but not both), we wouldn't be able to represent this in an openapi spec.

All of that makes me lean toward the current solution, but no particularly strongly. Do you have a preference, @lgalabru?

lgalabru: > All of that makes me lean toward the current solution

Agreed!

lgalabru: > All of that makes me lean toward the current solution

Agreed!

PR #382, opened 2023-08-11
update generated config with additional comments

ryanwaits:
### Description

update comments in the generated `Chainhook.toml` file around the `stacks_node_rpc_url` and `bitcoind_zmq_url` to add a bit more clarity around what they are and when to use one over the other.

### Checklist

- [ ] All tests pass


components/chainhook-cli/src/config/generator.rs
```diff
@@ -19,12 +19,15 @@ mode = "{network}"
...
 bitcoind_rpc_password = "devnet"
-# Bitcoin block events can be received by Chainhook
-# either through a Bitcoin node's ZeroMQ interface,
-# or through the Stacks node. The Stacks node is
-# used by default:
+
+# Chainhook must be able to receive Bitcoin block events.
+# These events can originate from either a Stacks node or a Bitcoin node's ZeroMQ interface.
+
+# By default, the service is set to receive Bitcoin block events from the Stacks node (via burnchain):
```
mefrem, line 26: "via burnchain" might beg some context... Burnchain is a deprecated PoX v1 term that persists in old Stacks Node software. Not even some (a lot of?) Stacks users might know what it is. Is there a better way to put it?

mefrem:
Made a comment, otherwise looks great! This is exactly what @MicaiahReid thought would make the docs stronger is providing more context to developers about "what is needed and when".

components/chainhook-cli/src/config/generator.rs
```diff
@@ -36,6 +39,9 @@ max_number_of_processing_threads = 16
 max_number_of_networking_threads = 16
 max_caching_memory_size_mb = 32000
 
+# If you don't require historical data for your predicates or wish to avoid downloading
```
MicaiahReid, line 42: The TSV file is not downloaded every time. It is downloaded once, and future runs use that cached version until a new one is available.

components/chainhook-cli/src/config/generator.rs
```diff
@@ -36,6 +39,8 @@ max_number_of_processing_threads = 16
 max_number_of_networking_threads = 16
 max_caching_memory_size_mb = 32000
 
+# The TSV file is required for downloading historical data for your predicates. If this is not a requirement, you can comment out the `tsv_file_url` line.
+#
```
MicaiahReid, line 43: Sorry for the nit! But this line was pretty long and we had an extra `#` at the end.
```suggestion
# The TSV file is required for downloading historical data for your predicates. 
# If this is not a requirement, you can comment out the `tsv_file_url` line.
```

PR #383, opened 2023-08-14
Quick tweaks

lgalabru:
- Bump retries counts for initial scans
- Introduce new ordinals curse type wildcard
- Simplify ordinals schemas by removing curse variants
- Bump versions
- Cargo format

### DIFF
--- a/components/chainhook-cli/src/scan/bitcoin.rs
+++ b/components/chainhook-cli/src/scan/bitcoin.rs
@@ -219,7 +219,7 @@ pub async fn execute_predicates_action<'a>(
                 actions_triggered += 1;
                 match action {
                     BitcoinChainhookOccurrence::Http(request) => {
-                        send_request(request, 10, 3, &ctx).await
+                        send_request(request, 10, 3, &ctx).await?
                     }
                     BitcoinChainhookOccurrence::File(path, bytes) => {
                         file_append(path, bytes, &ctx)?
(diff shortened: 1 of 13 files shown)


components/chainhook-cli/src/scan/bitcoin.rs
```diff
@@ -219,7 +219,7 @@ pub async fn execute_predicates_action<'a>(
                 actions_triggered += 1;
                 match action {
                     BitcoinChainhookOccurrence::Http(request) => {
-                        send_request(request, 3, 1, &ctx).await?
+                        send_request(request, 10, 3, &ctx).await
                     }
                     BitcoinChainhookOccurrence::File(path, bytes) => {
                         file_append(path, bytes, &ctx)?
```
MicaiahReid, line 225: The above (line 222) removed a `?`. Was that on purpose? If so, this line needs it removed as well, there is a mismatched return type on this match.

MicaiahReid (changes requested):

components/chainhook-cli/src/scan/bitcoin.rs
lgalabru: Thank you for catching these snafus!
Addressed with latest.

components/chainhook-sdk/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "chainhook-sdk"
-version = "0.8.0"
+version = "0.8.2"
```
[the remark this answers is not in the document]
lgalabru: Thank you for catching these snafus!
Addressed with latest.

MicaiahReid (approved):

PR #391, opened 2023-08-17
fix: use prevout to determine if tx is coinbase tx

MicaiahReid:


lgalabru:
Hey @MicaiahReid! thanks for the ping.
The `prevout` approach should work, but the `txid` test should have work too. Could you remind me what was the `txid` value we were logging?

MicaiahReid:
@lgalabru I actually think we (or I, haha) misunderstood the issue. I was getting the error: `Unable to standardize block#801221 00000000000000000002210b020a96cc3933866500b9e5b81da28ffa8d04c87c: error retrieving prevout for transaction 119afd239717e96f02d137bc6b724605f205e39de7286e5e16f397301d23ded6, input #0 (block #801221)` for this bitcoin mainnet predicate: 
```
"if_this": {
    "scope": "txid",
    "equals": "0x55c86631e9b08b817e6ebac00b9818d1fd49e0a7d6c00429b6e364d3ef845920"
}
```

Because `prevout` was not available on the transaction, we thought that a coinbase transaction was somehow making it past the `is_coinbase` check, which would mean the `txid.is_none()` check was insufficient.

The real problem, pointed out by @qustavo, is that before bitcoind [v23.0](https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-23.0.md#updated-rpcs), the `prevout` field was not included in the `getblock` RPC call. I was running bitcoind v22 before when I got this error. Upgrading to v25 fixes this.

Do we call this a noop? Or do we need to support previous bitcoind versions?

MicaiahReid:
I think we are settling on requiring bitcoind v24+ to use Chainhook, so closing this PR

PR #393, opened 2023-08-17
fix: retrieve blocks until tip

qustavo:
### Description
When scanning the chain `chainhook predicates scan` setting a `start_block` predicate param, blockchain tip is not evaluated. I don't know if this was intended (leaving tip unscanned) but it's misleading to skip the tip.

#### Breaking change?

Unlikely


lgalabru (approved):
Thanks for catching this issue @qustavo!

PR #395, opened 2023-08-21
fix: zmq reorg handler

lgalabru:
As recently experienced, the zmq channel can behave differently in presence of re-orgs.
This PR is tweaking the zmq code site to support the following scenario. 
 ```
  // Handle a behaviour specific to ZMQ usage in bitcoind.
  // Considering a simple re-org:
  // A (1) - B1 (2) - C1 (3)
  //       \ B2 (4) - C2 (5) - D2 (6)
  // When D2 is being discovered (making A -> B2 -> C2 -> D2 the new canonical fork)
  // it looks like ZMQ is only publishing D2.
  // Without additional operation, we end up with a block that that we
  // can't append.
```

The reorg handler can support blocks out of order and these situations are pretty well unit tested, but in this situation, we never received the missing block.

We are also revisiting the anatomy of the integration with Ordhook, by introducing a new `ObserverSidecar` struct providing more context to layers built on top of Chainhook.


MicaiahReid (approved):
I don't understand how this fixes the zmq reorg handler, but I'm cool with the changes 😅

MicaiahReid (approved):

Issue #298, opened 2023-06-27
keep predicate registration api open when startup predicates are passed


MicaiahReid:
Currently, if `chainhook service start` is used with the `--predicate-path` argument, we don't start the predicate registration API, even if it's enabled in the config:
https://github.com/hirosystems/chainhook/blob/69e122112740e5e8c9f3f468c49b1fe905f22f69/components/chainhook-cli/src/cli/mod.rs#L440

We should probably start the API regardless of if predicates are passed on startup.

sabbyanandan:
Agreed on starting the API regardless of the predicates.

PR #397, opened 2023-08-24
feat: augment predicate status returned by GET/LIST endpoints

### LINKED ISSUE #324: Add last triggered timestamp to GET and LIST endpoints
Aug 21st, 2023:
After discussing this requirement with @MicaiahReid and @BLuEScioN, we will add `last_triggered` (i.e., timestamp) as part of the GET and LIST endpoints.
Sabby will open a separate issue to account for the error messages.

Was 2nd:

>Add these fields to the `GET` and `LIST` endpoint:
1: Last Updated: Timestamp
2: Last Triggered: Timestamp
>Add this to `GET` endpoint:
1: 1-line of the last log: String (maybe 100 chars?)

Was 1st: 
> Currently its only possible to tell whether a chainhook is still active by determining whether the endblock specified in the chainhook config has been surpassed or not, or whether the startblock has been reached yet. But there is a third bit of information that may determine whether or not the chainhook is still active - the number of occurences before expiration. Currently, there is no way to retrieve how many times the chainhook has been activated. Thus, there is no way to accurately determine if the chainhook is still active. I recommend that an isActive prop be added to the GET and/or LIST endpoints to make it easier to consume this information. This kind of information would be very useful to display for chainhooks on the Platform.

### LINKED ISSUE #390: `chainhook service` does not register predicate when using --predicate-path flag
**Describe the bug**
When running `chainhook service` using the `--predicate-path` flag, the predicate does not get registered and the expected events are not triggered.

**To Reproduce**
[here](https://www.loom.com/share/fb831a660d9b4393b9b7977ddf1a228b) is a loom going over a working example without the predicate path flag, and one without

**Expected behavior**
When running the `chainhook service` using the predicate path flag, the predicate should be registered automatically and begin successfully triggering events based on the predicate config, similar to the two step process of running the service and then registering the predicate.

### LINKED ISSUE #396: Include error-message as part of the GET/LIST endpoints to help with troubleshooting
As a user, when the actively observing chainhooks fail to trigger either because of the errors in Chainhook itself or due to the issues when triggering the HTTP endpoint for a given event, I would like to know what and why it failed, so I can troubleshoot and course-correct.

The interruptions could be:
a: Errors in Chainhook
b: Having an incorrect HTTP endpoint as a trigger
c: Having an unresponsive or erroring HTTP endpoint as a trigger
d: Any runtime errors between the Chainhook node, Chainhook API, or HTTP endpoints

If we can include the error-message and can define a category for those errors, the Hiro Platform can handle them and gracefully let users know of the problem.

### LINKED ISSUE #402: Restarting predicate server mid-scan does not continue scan

### LINKED ISSUE #403: Restarting predicate server does check blocks mined between restarts
If you have predicates in `streaming` status, being checked against new blocks, then you stop the server, wait for blocks to be mined, then restart the server, the blocks that were mined since terminating are not scanned against those streaming predicates

MicaiahReid:
### Description

This PR accomplishes a few things:
 - renames/adds/removes predicate statuses according to the flowchart in docs/images/predicate-status-flowchart/PredicateStatusFlowchart.png
 - adds more context to errors returned in Interrupted status
 - adds status data to `GET /v1/chainhooks` endpoint
 - fixes bug in chainhook service where block streaming continued past `end_block`

Fixes #396, fixes #324 

Also: 
Improves how we handle a restart of `chainhook service` while predicates are scanning/streaming. Here are the cases we now handle:
1. Predicates that were in `scanning` status when Chainhook was terminated will resume scanning starting from their `last_evaluated_block_height`. *Note: because we only save predicate status every 10 scans, we could end up re-emiting matches on a resetart*
2. Predicates that were in `new` status when Chainhook was terminated will start scanning at the predicate's `start_block`
3. Predicates that were in `streaming` status will _return_ to a `scanning` status, starting at `last_evaluated_block_height` to catch up on the missed blocks. Note, the `number_of_blocks_to_scan` is set to 0 for this temporary catch-up, as it's difficult to compute the number of remaining blocks in the context of this change
4. If predicates were passed in at startup, we also register those to begin scanning, which previously didn't happen
5. We now allow passing in a predicate at startup _and_ registering additional predicates with the predicate registration server. This means that if you use the same startup predicate repeatedly, it will already be saved in redis and _not_ be reloaded.

Fixes: #298, fixes #390, fixes #402, fixes #403
#### Breaking change?

The rename of `ScanningData`'s `number_of_blocks_sent` field could technically be considered breaking, let's discuss.

### Checklist

- [x] All tests pass
Test coverage before: 23.2%
Test coverage after: 37.72%

### DIFF
--- a/dockerfiles/components/chainhook-node.dockerfile
+++ b/dockerfiles/components/chainhook-node.dockerfile
@@ -4,7 +4,7 @@ WORKDIR /src
 
 RUN apt update && apt install -y ca-certificates pkg-config libssl-dev libclang-11-dev
 
-RUN rustup update 1.67.0 && rustup default 1.67.0
+RUN rustup update 1.70.0 && rustup default 1.70.0
 
 COPY ./components/chainhook-cli /src/components/chainhook-cli
 


components/chainhook-cli/src/service/mod.rs
```diff
@@ -345,23 +366,187 @@ impl Service {
...
     Interrupted(String),
     Disabled,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
 pub struct ScanningData {
     pub number_of_blocks_to_scan: u64,
     pub number_of_blocks_scanned: u64,
-    pub number_of_blocks_sent: u64,
```
MicaiahReid, line 357: It seemed that Sabby had a misunderstanding of what `number_of_blocks_sent` means, so I thought a rename was in order. I'm open to suggestions other than `number_of_times_triggered`

components/chainhook-cli/src/scan/stacks.rs
```diff
@@ -201,10 +198,10 @@ pub async fn scan_stacks_chainstate_via_rocksdb_using_predicate(
     let mut err_count = 0;
     let number_of_blocks_to_scan = block_heights_to_scan.len() as u64;
     let mut number_of_blocks_scanned = 0;
-    let mut number_of_blocks_sent = 0u64;
+    let mut number_of_times_triggered = 0u64;
 
     while let Some(current_block_height) = block_heights_to_scan.pop_front() {
-        number_of_blocks_scanned += 1;
+        number_of_blocks_scanned += 1; // todo: can we remove this and just use `blocks_scanned`?
```
MicaiahReid, line 204: @lgalabru just wanted a sanity check - these are declared a few lines apart and are incremented a few lines apart under the same circumstances - is there a reason for them to be separate?

components/chainhook-cli/src/scan/stacks.rs
```diff
@@ -245,40 +242,55 @@ pub async fn scan_stacks_chainstate_via_rocksdb_using_predicate(
             apply: hits_per_blocks,
             rollback: vec![],
         };
-        match handle_stacks_hook_action(trigger, &proofs, &ctx) {
+        let res = match handle_stacks_hook_action(trigger, &proofs, &ctx) {
             Err(e) => {
                 error!(ctx.expect_logger(), "unable to handle action {}", e);
+                Ok(()) // todo: should this error increment our err_count?
```
MicaiahReid, line 248: @lgalabru, another sanity check - previously we were just logging this error. Do we want this error to possibly get propagated to the user via an `Interrupt` status?

components/chainhook-sdk/src/observer/mod.rs
```diff
@@ -1122,6 +1140,11 @@ pub async fn start_observer_commands_handler(
                     let mut total_occurrences: u64 = *chainhooks_occurrences_tracker
                         .get(&trigger.chainhook.uuid)
                         .unwrap_or(&0);
+                    // todo: this currently is only additive, and an occurrence means we match a chain event,
+                    // rather than the number of blocks. Should we instead add to the total occurrences for
+                    // every apply block, and subtract for every rollback? If we did this, we could set the
+                    // status to `Expired` when we go above `expire_after_occurrence` occurrences, rather than
+                    // deregistering
                     total_occurrences += 1;
```
MicaiahReid, line 1148: @lgalabru I can either make an issue for this if you think it makes sense, or remove the todo

components/chainhook-cli/src/scan/stacks.rs
lgalabru: you mean updating redis?
that state machine might become a bit tricky to handle but if you're confident and if we have tests, let's do it.

components/chainhook-cli/src/service/mod.rs
```diff
@@ -509,18 +601,21 @@ fn set_predicate_streaming_status(
                     number_of_times_triggered,
                     last_evaluated_block_height,
                 ),
-                PredicateStatus::Expired(ExpiredData {
+                PredicateStatus::ExpiredUnsafe(ExpiredData {
```
lgalabru, line 604: `Unsafe` could be ambiguous, how about `Expired` + `UnconfirmedExpiration` or `ConfirmedExpiration` + `Expired`?

components/chainhook-sdk/src/chainhooks/types.rs
```diff
@@ -380,6 +406,7 @@ impl StacksChainhookFullSpecification {
             predicate: spec.predicate,
             action: spec.action,
             enabled: false,
+            expired: false,
```
lgalabru, line 409: Is the state `enabled = true, expired = true` reachable?

MicaiahReid: No, it isn't. Do you want to use just `enabled`?

components/chainhook-cli/src/scan/stacks.rs
MicaiahReid: Created #412

components/chainhook-sdk/src/observer/mod.rs
MicaiahReid: Created #413

components/chainhook-cli/src/service/mod.rs
MicaiahReid: I like it. I went with`UnconfirmedExpiration` +`ConfirmedExpiration` just so they can be consistent and explicit.

components/chainhook-sdk/src/chainhooks/types.rs
lgalabru: I am not sure, it feels like that `expired` is redundant, but we need some information.
Could 
```rust
/// Block height
expired_at: Option<u64>
```
be useful?

MicaiahReid: Changed: https://github.com/hirosystems/chainhook/pull/397/commits/f6970c7075cd623041ff348f74213abacfc813c5

lgalabru (approved):
Thanks @MicaiahReid!

PR #408, opened 2023-09-04
fix: prevent panic when scanning from genesis block

qustavo:
Parent block height (`u64`) is always set to `block - 1` which will result on a panic when trying to construct genesis parent's block height.


lgalabru (approved):
Thanks @qustavo!

MicaiahReid:
Thanks, @qustavo! There are other places in the codebase we do this. If you search ` - 1` throughout the codebase, you'll see multiple times where we use this same logic to calculate the parent block height. Can you fix these as well?

Let me know if you don't have the bandwidth for this and I can take it on eventually.

qustavo:
@MicaiahReid fixed.
This PR does not touch the code affecting microblocks, although incorrect, micoblock number starts with 1, so u64 never gets a negative value.

MicaiahReid (approved):

PR #415, opened 2023-09-15
test: add tests for `GET /v1/chainhooks/<uuid>` route

MicaiahReid:
### Description

Add tests for chainhook service, fix a few things along the way.

---

### Checklist

- [X] All tests pass
- [X] Tests added in this PR (if applicable)
Test coverage before: 37.72%
Test coverage after: 47.07%


components/chainhook-cli/src/service/tests/mod.rs
```diff
@@ -559,89 +271,442 @@ fn assert_streaming_status(status: PredicateStatus) {
...
+        logger: Some(logger),
+        tracer: false,
+    };
+
+    write_stacks_blocks_to_tsv(starting_chain_tip, &tsv_dir).unwrap_or_else(|e| {
+        std::fs::remove_dir_all(&working_dir).unwrap();
+        flush_redis(redis_port);
+        redis_process.kill().unwrap();
+        panic!("test failed with error: {e}");
+    });
```
MicaiahReid, line 348: @lgalabru  I repeatedly use this pattern a lot, but I'd really like to be able to use something like: 

```suggestion
    write_stacks_blocks_to_tsv(starting_chain_tip, &tsv_dir).map_err(|e| cleanup(e.to_string(), redis_process,redis_port,working_dir))?;
```

My problem is that passing `redis_process` into a function moves it, so it can't be called by `cleanup` more than once. This doesn't really matter - cleanup always panics and can only be called once. However, the borrow checker doesn't know this. 

The obvious workaround is to use `redis_process.clone()`, but its type is `std::process::Child`, which we obviously cannot alter and doesn't have `Clone` implemented.

Do you know of a way that I can simplify this pattern and reduce the number of lines of code? Maybe we could pair on this.

components/chainhook-sdk/src/indexer/bitcoin/mod.rs
```diff
@@ -575,7 +578,7 @@ fn try_parse_stacks_operation(
 
     // Safely parsing the first 2 bytes (following OP_RETURN + PUSH_DATA)
     let op_return_output = &outputs[0].script_pub_key.hex;
-    if op_return_output.len() < 7 {
+    if op_return_output.len() < CONFIRMED_SEGMENT_MINIMUM_LENGTH as usize {
```
lgalabru, line 581: 👍

lgalabru (approved):
This PR looks great, thank you so much @MicaiahReid!

PR #417, opened 2023-09-18
fix: add event index to transaction events

### LINKED ISSUE #387: return `event_index` in Stacks transaction event metadata
Stacks transaction events have an event index that indicates their ordering inside it. They are sent from the Stacks node to RPC listeners like so (see `event_index`):
```json
    {
      "txid": "0xb55e53065b3e934716eb0f7dcce5138b719ea247459f18f60893e15c8d9a0159",
      "type": "ft_mint_event",
      "committed": true,
      "event_index": 46,
      "ft_mint_event": {
        "amount": "40000000000000",
        "recipient": "SP16KWQY6ZPXNYT43A4RKXBXMFT3V7ZMV5YYNR1CK",
        "asset_identifier": "SP3XYJ8XFZYF7MD86QQ5EF8HBVHHZFFQ9HM6SPJNQ.meme-token::MEME"
      }
    },
```
However, chainhook payloads do not return these indexes inside events, which makes it impossible to sort them by the order in which they were processed (no `event_index` here):
```json
{
  "data": {
    "amount": "144",
    "asset_identifier": "SP466FNC0P7JWTNM2R9T199QRZN1MYEDTAR0KP27.miamicoin-token::miamicoin",
    "recipient": "SPCXZRY4FQZQHQ1EMBSDM1183HDK2ZGHPW3M1MA4",
    "sender": "SP3HXJJMJQ06GNAZ8XWDN1QM48JEDC6PP6W3YZPZJ"
  },
  "type": "FTTransferEvent"
}
```

Could this index be returned with these events? This is necessary for the Token Metadata API to process notifications correctly. Ideally, events should look like this:
```json
{
  "data": {
    "amount": "144",
    "asset_identifier": "SP466FNC0P7JWTNM2R9T199QRZN1MYEDTAR0KP27.miamicoin-token::miamicoin",
    "recipient": "SPCXZRY4FQZQHQ1EMBSDM1183HDK2ZGHPW3M1MA4",
    "sender": "SP3HXJJMJQ06GNAZ8XWDN1QM48JEDC6PP6W3YZPZJ"
  },
  "type": "FTTransferEvent",
  "position": {
    "index": 46,
  }
}
```

janniks:
- closes #387 
- adds a position struct to the event and renames the event to payload (with enum type)
- updates occurences.json to include the expected serialized position for testing


MicaiahReid (approved):
Thanks, @janniks!!

MicaiahReid (changes requested):
I've discovered some ways where this may cause breakages, I need to do some investigation.

MicaiahReid (approved):
Okay, it looks like it should be good!

lgalabru:
Thanks @janniks! We already have the struc `StacksTransactionPosition` in place, indicating the position of a transaction in the block or in the microblock, could we use this instead?
Apologies for the latency on this review.

janniks:
Oh yeah sure, sorry for missing that 👍

janniks:
The new positions are referring to events not blocks. Should we create an enum type that mixed blocks and events (might be funky when matching. Or add an additional enum layer EventPosition BlockPosition (but that also sounds complected)

MicaiahReid:
@lgalabru I don't know if I agree - Jannik's change is adding a struct for an event's position in a transaction, while `StacksTransactionPosition` represents a transaction's position within a block. 

I guess we could abstract that position enum to be more general, from:
```rust
pub enum StacksTransactionPosition {
    AnchorBlock(AnchorBlockPosition),
    MicroBlock(MicroBlockPosition),
}
```
to:
```rust
pub enum Position {
    TransactionPositionInAnchorBlock(AnchorBlockPosition),
    TransactionPositionInMicroBlock(MicroBlockPosition),
    EventPositionInTransaction(u64)
}
```

but that feel's clunky. Let me know if we're misunderstanding what you meant here!

lgalabru (approved):
My apologies for the latency on this PR, yes that makes sense to me.
Thank you for your help @janniks!

PR #421, opened 2023-09-29
ci: update CD process

deantchi:
changes to the CI workflow.
adding auto deployments using docker container digest to the k8s repo.
dev and stg will auto approve, while prod will require manual approval

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -136,8 +136,6 @@ jobs:
         uses: ./actions/deploy
         with:
           docker_tag: ${{ needs.build-publish.outputs.docker_image_digest }}
-          k8s_repo: k8s
-          k8s_branch: next
           file_pattern: manifests/chainhooks/${{ matrix.k8s-env }}/chainhook-node/${{ env.DEPLOY_ENV }}/base/kustomization.yaml
           gh_token: ${{ secrets.GH_TOKEN }}
 
@@ -180,8 +178,6 @@ jobs:
         uses: ./actions/deploy
         with:
           docker_tag: ${{ needs.build-publish.outputs.docker_image_digest }}
-          k8s_repo: k8s
-          k8s_branch: next
           file_pattern: manifests/chainhooks/${{ matrix.k8s-env }}/chainhook-node/${{ env.DEPLOY_ENV }}/base/kustomization.yaml
           gh_token: ${{ secrets.GH_TOKEN }}
 
@@ -226,7 +222,5 @@ jobs:
         uses: ./actions/deploy
         with:
           docker_tag: ${{ needs.build-publish.outputs.docker_image_digest }}
-          k8s_repo: k8s
-          k8s_branch: next
           file_pattern: manifests/chainhooks/${{ matrix.k8s-env }}/chainhook-node/${{ env.DEPLOY_ENV }}/base/kustomization.yaml
           gh_token: ${{ secrets.GH_TOKEN }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -120,7 +120,7 @@ jobs:
       DEPLOY_ENV: dev
       DEPLOY_NETWORK: mainnet
     environment:
-      name: Development-${{ env.DEPLOY_NETWORK }}
+      name: Development-mainnet
       url: https://platform.dev.hiro.so/
     steps:
       - name: Checkout actions repo
@@ -148,7 +148,7 @@ jobs:
       DEPLOY_ENV: dev
       DEPLOY_NETWORK: testnet
     environment:
-      name: Development-${{ env.DEPLOY_NETWORK }}
+      name: Development-testnet
       url: https://platform.dev.hiro.so/
     steps:
       - name: Checkout actions repo
@@ -191,7 +191,7 @@ jobs:
       DEPLOY_ENV: stg
       DEPLOY_NETWORK: mainnet
     environment:
-      name: Staging-${{ env.DEPLOY_NETWORK }}
+      name: Staging-mainnet
       url: https://platform.stg.hiro.so/
     steps:
       - name: Checkout actions repo
@@ -220,7 +220,7 @@ jobs:
       DEPLOY_ENV: stg
       DEPLOY_NETWORK: testnet
     environment:
-      name: Staging-${{ env.DEPLOY_NETWORK }}
+      name: Staging-testnet
       url: https://platform.stg.hiro.so/
     steps:
       - name: Checkout actions repo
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -77,3 +77,109 @@ jobs:
...
+
+  devploy-dev:
+    runs-on: ubuntu-latest
+    needs:
+      - build-publish
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
+    env:
+      DEPLOY_ENV: dev
+    environment:
+      name: Development
```
CharlieC3, line 89: These 3 envs need to have required reviewers configured, otherwise it'll just deploy to them without an approval. [You can configure that here.](https://github.com/hirosystems/chainhook/settings/environments)

.github/workflows/ci.yaml
```diff
@@ -77,3 +77,109 @@ jobs:
...
+        uses: actions/checkout@v4
+        with:
+          ref: main
+          token: ${{ secrets.GH_TOKEN }}
+          repository: ${{ secrets.DEVOPS_ACTIONS_REPO }}
+
+      - name: Deploy Chainhook build to Dev
+        uses: ./actions/deploy
+        with:
+          docker_tag: ${{ needs.build-publish.outputs.docker_image_digest }}
```
CharlieC3, line 101: I don't believe the build-publish job is outputting anything, so this will fail. Double check any other outputs used too.

.github/workflows/ci.yaml
```diff
@@ -77,3 +77,109 @@ jobs:
...
+          ENV_ID=$(curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" -H "Accept: application/vnd.github.v3+json" "https://api.github.com/repos/hirosystems/chainhook/actions/runs/${{ github.run_id }}/pending_deployments" | jq -r '.[0].environment.id // empty')
+          if [[ -n "${ENV_ID}" ]]; then
+            curl -s -X POST -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" -H "Accept: application/vnd.github.v3+json" "https://api.github.com/repos/hirosystems/chainhook/actions/runs/${{ github.run_id }}/pending_deployments" -d "{\"environment_ids\":[${ENV_ID}],\"state\":\"approved\",\"comment\":\"auto approve\"}"
+          fi
+
+  deploy-prod:
+    runs-on: ubuntu-latest
+    needs:
+      - build-publish
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
```
CharlieC3, line 165: This doesn't seem right, we don't want to allow deployments to prod except under more rigid conditions. Double check this condition with the workflow in the [stacks-devnet-api](https://github.com/hirosystems/stacks-devnet-api/blob/develop/.github/workflows/ci.yaml#L287).

CharlieC3 (changes requested):
Good start! Just a few tweaks.

.github/workflows/ci.yaml
deantchi: opps yeah, I added the reviewers but forgot to save the changes

CharlieC3:
@deantchi you'll also need to update the chainhook manifests to apply the environment-level patch using `digest` instead of `newTag`.
See the Explorer or stacks-devnet-api manifests for a reference if needed.

.github/workflows/ci.yaml
```diff
@@ -77,3 +77,109 @@ jobs:
...
+          token: ${{ secrets.GH_TOKEN }}
+          repository: ${{ secrets.DEVOPS_ACTIONS_REPO }}
+
+      - name: Deploy Chainhook build to Prd
+        uses: ./actions/deploy
+        with:
+          docker_tag: ${{ needs.build-publish.outputs.docker_image_digest }}
+          k8s_repo: k8s
+          k8s_branch: next
+          file_pattern: manifests/chainhooks/mainnet/chainhook-node/${{ env.DEPLOY_ENV }}/base/kustomization.yaml
```
CharlieC3, line 184: We should take testnet into account as well I think.

deantchi: trying out the use of matrix strategy for testnet | mainnet

deantchi: this seems to be working.  https://github.com/hirosystems/chainhook/actions/runs/6356417456

.github/workflows/ci.yaml
```diff
@@ -70,10 +73,129 @@ jobs:
...
+
+  deploy-prod:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        k8s-env: [mainnet,testnet]
+    needs:
+      - build-publish
+      - deploy-staging
+    if: needs.build-publish-release.outputs.new_release_published == 'true' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
```
CharlieC3, line 181:
```suggestion
    if: needs.build-publish.outputs.new_release_published == 'true' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
```

.github/workflows/ci.yaml
```diff
@@ -70,10 +73,129 @@ jobs:
...
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           file: ./dockerfiles/components/chainhook-node.dockerfile
           # Only push if (there's a new release on main branch, or if building a non-main branch) and (Only run on non-PR events or only PRs that aren't from forks)
           push: ${{ (github.ref != 'refs/heads/master' || steps.semantic.outputs.new_release_version != '') && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository) }}
+
+  deploy-dev:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
CharlieC3, line 88: I like what you did with the matrix! However on the approval screen, it doesn't seem like it's going to let us differentiate between a testnet dev deployment and a mainnet dev deployment. So I think they only way to get around this is to create a total of 6 Github Environments, aptly named something like Development - Mainnet, Development - Testnet, etc. You might still be able to leverage the matrix usage here as well.

CharlieC3 (changes requested):

.github/workflows/ci.yaml
CharlieC3: Despite there being two deployments prepped, there's only one approval box. Ideally devs would have control to deploy to both networks for an environment, or just one.
Screenshot 2023-09-29 at 8 39 30 PM

lgalabru:
🤤 
Thank you for this new flow @deantchi @CharlieC3!

MicaiahReid (approved):

.github/workflows/ci.yaml
deantchi: created 6 separate deployment environments as suggested and using matrix function

CharlieC3 (approved):
I made a few changes to help with CI performance, specifically added caching for cargo tests and docker image build. In addition to that, I fixed some dependency issues between jobs, ensured the new tag is being checked out once a new release is published, removed the `blockstack` Docker hub org from being used, and some other minor fixes.

I'd like either @MicaiahReid or @deantchi to check this PR again and re-approve if possible.

PR #425, opened 2023-10-03
fix: make predicate status timestamps optional

MicaiahReid:
### Description

This changes the predicate status timestamps:
 - from String (converted from a u128) to a u64
   - this means we are now storing the data in seconds rather than milliseconds
 - makes them optional (so they will not appear in a serialized status if set to `None`)

#### Breaking change?

Previously the timestamp was in milliseconds, now it's in seconds. Anyone depending on the timestamp being in milliseconds could be impacted by this change.

---

### Checklist

- [ ] All tests pass


lgalabru (approved):
Thanks @MicaiahReid!

PR #430, opened 2023-10-04
fix: add nonce to serialized stacks tx

### LINKED ISSUE #422: Include "nonce" as part of the event stream response
As a user, I want to use "stx-address" along with the transaction's "nonce" as a combination to prove/confirm that something is complete. To do that, I am using Chainhook to observe transaction events and are expecting the "nonce" as part of the event-streams. 

Using the Stacks Blockchain API to get the latest "nonce" via https://docs.hiro.so/api/get-the-latest-nonce-used-by-an-account is certainly an option; however, I could avoid an extra call.

Requested by: https://zeroauthority.xyz/

MicaiahReid:
Fixes #422

### DIFF
--- a/components/chainhook-sdk/src/chainhooks/stacks/mod.rs
+++ b/components/chainhook-sdk/src/chainhooks/stacks/mod.rs
@@ -544,6 +544,7 @@ fn serialize_stacks_transaction(
                 json!(transaction.metadata.result)
             },
             "sender": transaction.metadata.sender,
+            "nonce": transaction.metadata.nonce,
             "fee": transaction.metadata.fee,
             "kind": transaction.metadata.kind,
             "receipt": {
--- a/components/chainhook-sdk/src/chainhooks/tests/fixtures/stacks/testnet/occurrence.json
+++ b/components/chainhook-sdk/src/chainhooks/tests/fixtures/stacks/testnet/occurrence.json
@@ -45,6 +45,7 @@
               },
               "type": "ContractCall"
             },
+            "nonce": 4064,
             "position": {
               "index": 1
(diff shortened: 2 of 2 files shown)


components/chainhook-sdk/src/chainhooks/stacks/mod.rs
```diff
@@ -544,6 +544,7 @@ fn serialize_stacks_transaction(
                 json!(transaction.metadata.result)
             },
             "sender": transaction.metadata.sender,
+            "nonce": transaction.metadata.nonce,
```
sabbyanandan, line 547: nice; glad it was as straightforward as this!

I see references to ["hex string"](https://github.com/hirosystems/chainhook/blob/ec76f29dd95818f2645f04d61ebaf3a8af1bad87/components/chainhook-types-rs/src/rosetta.rs#L19), so how do we feel about this feedback/ask: https://github.com/hirosystems/chainhook/issues/422#issuecomment-1742285097?

sabbyanandan (changes requested):
Curious about the other ask from the Zero, which I linked in my comment.

lgalabru (approved):
Thanks @MicaiahReid!

components/chainhook-sdk/src/chainhooks/stacks/mod.rs
MicaiahReid: We need further clarification on that. I can work it in a separate PR once we have the info we need. Also, I've created a separate issue to track it: #431

sabbyanandan (approved):
Heck yeah; thank you!

PR #432, opened 2023-10-05
fix: make `include_contract_abi` optional

MicaiahReid:
### Description

The new field `include_contract_abi` already was optional for the user-facing type `StacksChainhookNetworkSpecification`, but was not optional for the internal type `StacksChainhookSpecification`. Any time we converted to a `StacksChainhookSpecification`, we'd `.unwrap_or(false)`, which seemed okay.

However, a running Chainhook node has the `StacksChainhookSpecification` type saved in redis, so whenever a pre-existing node is restarted after upgrading, it is missing the required `include_contract_abi` field and errors.

This PR makes the field optional everywhere.

I'll need to think about how to add tests to prevent this sort of bug in the future.

### Checklist

- [x] All tests pass


lgalabru (approved):
LGTM! thanks @MicaiahReid!

PR #442, opened 2023-10-11
fix: redis conn

lgalabru:
This PR is fixing the management of the redis connection. In the current approach, we're reusing an existing connection that is being opened at startup for test purposes, but that can get broken after some time if blocks take time to arrive.
Fix: we open/close a new connection every time a new block hits the API.


MicaiahReid (approved):
🎉

PR #449, opened 2023-10-12
fix(stacks-indexer): prevent subtract with overflow

MicaiahReid:
I was attempting to update Clarinet to the latest version of chainhook-sdk/types, and I ran into this issue when testing the devnet.

The devnet mines 100 bitcoin blocks, and the default `pox_info.first_burnchain_block_height` is also 100, so devnet startup would always fail.

### DIFF
--- a/components/chainhook-sdk/src/indexer/stacks/mod.rs
+++ b/components/chainhook-sdk/src/indexer/stacks/mod.rs
@@ -253,8 +253,10 @@ pub fn standardize_stacks_block(
     let pox_cycle_length: u64 = (chain_ctx.pox_info.prepare_phase_block_length
         + chain_ctx.pox_info.reward_phase_block_length)
         .into();
-    let current_len =
-        block.burn_block_height - (1 + chain_ctx.pox_info.first_burnchain_block_height);
+    let current_len = u64::saturating_sub(
+        block.burn_block_height,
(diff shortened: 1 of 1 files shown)


components/chainhook-sdk/src/indexer/stacks/mod.rs
```diff
@@ -253,8 +253,12 @@ pub fn standardize_stacks_block(
     let pox_cycle_length: u64 = (chain_ctx.pox_info.prepare_phase_block_length
         + chain_ctx.pox_info.reward_phase_block_length)
         .into();
-    let current_len =
-        block.burn_block_height - (1 + chain_ctx.pox_info.first_burnchain_block_height);
+    let current_len = if chain_ctx.pox_info.first_burnchain_block_height >= block.burn_block_height
+    {
+        0
+    } else {
+        block.burn_block_height - (1 + chain_ctx.pox_info.first_burnchain_block_height)
```
lgalabru, line 260: Thanks for catching this @MicaiahReid! could we use `saturating_sub` instead?

MicaiahReid: Nice, I didn't know about that! Done!

lgalabru (approved):
Thanks!

PR #450, opened 2023-10-13
fix: buffer decoding of archive file to reduce memory usage

### LINKED ISSUE #401: Optimize memory at boot when ingesting tsv events
In the current approach, the whole archive file (10Gb+) is loaded in memory, which is something we should optimize.

MicaiahReid:
Fixes #401 

Previously, memory usage peaked at 32 GB. Now, I didn't see it go above 8 MB


lgalabru (approved):
Thanks @MicaiahReid!

PR #457, opened 2023-11-01
feat: add inscription transfer destination schema

rafaelcr:


lgalabru (approved):
Looking good, thanks @rafaelcr!

PR #459, opened 2023-11-15
chore: ugprade dependencies from clarinet

hugocaillard:
### Description

Upgrade dependencies from Clarinet and make less strict.
This way, clarinet don't have to pull different version of clarinet-files, stacks-rpc-client, and sub-dependencies when pulling chainhook crates

Made some other changes to the dependency handling in `Cargo.toml`s


rust-toolchain
```diff
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.70.0"
+channel = "stable"
```
hugocaillard, line 2: Oops, didn't want to include this in this PR, I was curious to see if it build on stable (1.73.0), and it does.

I can remove if you think it's safer to stick to 1.70, or make the change in an other PR

components/chainhook-sdk/Cargo.toml
```diff
@@ -8,17 +8,17 @@ edition = "2021"
...
 serde_derive = "1"
-stacks-rpc-client = "1.0.10"
+stacks-rpc-client = "2"
 clarinet-utils = "1.0.0"
 hiro-system-kit = { version = "0.3.1", optional = true }
 # stacks-rpc-client = { version = "1", path = "../../../clarinet/components/stacks-rpc-client" }
 # clarinet-utils = { version = "1", path = "../../../clarinet/components/clarinet-utils" }
 # hiro-system-kit = { version = "0.1.0", path = "../../../clarinet/components/hiro-system-kit" }
-chainhook-types = { version = "1.1.0", path = "../chainhook-types-rs" }
+chainhook-types = { version = "1", path = "../chainhook-types-rs" }
```
hugocaillard, line 21: I find it easier to manage internal dependencies like so, let me know what you think

components/chainhook-cli/Cargo.toml
```diff
@@ -51,7 +53,6 @@ features = ["lz4", "snappy"]
 [dev-dependencies]
 criterion = "0.3"
 redis = "0.21.5"
-clarity-repl = "=1.7.1"
```
hugocaillard, line 54: Apparently not need anymore 🤔

lgalabru: ah yes, we're now re-exporting the repl in `stacks-rpc-client`, and using it from there to simplify dependencies.

rust-toolchain
lgalabru: `stable` would force us to be pro-active on testing this project against new toolchains as soon as they are released. I like the idea, but we might not have this bandwidth.

lgalabru: but it you tested with 1.73, let's upgrade!

components/chainhook-sdk/Cargo.toml
lgalabru: I don't think that the `chainhook-sdk` and `chainhook-types-rs` are mature enough to go with a loose specification. We're not strictly using semver yet, and sometimes just bumping the minor / patch for non backward compatible changes.
The interfaces of the SDK and types are still a bit in flux and being refined, could we please push these (reasonable) changes for later?

rust-toolchain
hugocaillard: Yeah indeed it might mean code breaking for now apparent reason in the future.
I switched to 1.73 then

components/chainhook-sdk/Cargo.toml
hugocaillard: Rollbacked

lgalabru (approved):
Thanks @hugocaillard!

PR #460, opened 2023-11-29
fix: update ordhook URLs on typescript client

rafaelcr:
Updates the TypeScript client URLs so they match with ordhook or chainhook requirements.


lgalabru (approved):
Thanks @rafaelcr!

Issue #466, opened 2023-12-22
implement more useful alerts on chainhook node health


lgalabru:
Now that the ping endpoint is reporting the bitcoin / stacks block heights, the alarming should be revisited, taking this into account.

lgalabru:
![Screenshot 2023-12-22 at 16 44 37](https://github.com/hirosystems/chainhook/assets/87777/8dc5d3c8-16aa-4624-9aa9-ade06e9790b6)

smcclellan:
@lgalabru to indicate which alarms are needed on this repo.

MicaiahReid:
Just adding a quick follow up here - I'm taking on this issue. We'll be adding alerts to Grafana that will notify us if Chainhook block ingestion falls behind block production for either the stacks or bitcoin nodes. 

This is requiring us to update Chainhook to emit Prometheus metrics. PR will be incoming soon!

MicaiahReid:
Closing, because everything is complete on the chainhook side with PR #473. The devops side is being tracked by https://github.com/hirosystems/devops/issues/1543

PR #473, opened 2024-01-10
feat: optionally serve Prometheus metrics

### LINKED ISSUE #474: backwards logic on `disabled` field for predicate api
When parsing the predicate api config file, we have some backwards logic on disabling:
https://github.com/hirosystems/chainhook/blob/823f4300c5b65ee006cdba1c6587fb549dcc1a33/components/chainhook-cli/src/config/mod.rs#L154-L162

If `disabled` is `true`, the api should be off.

MicaiahReid:
### Description

To enable improved alerts on downtime for Hiro's hosted Chainhook service, we need Chainhook to provide metrics that can be ingested by Prometheus. This PR changes some how we track our metrics (that are served over the `/ping` endpoint of the observer) to enable Prometheus compatibility, and adds a flag to optionally start a server to supply metrics to a Prometheus client.

### Example

Starting chainhook with the `--prometheus-port XXXX` flag now enables a service that can supply Prometheus metrics at `localhost:XXXX/metrics`.

If using a config file, this option can be specified via:
```yaml
[monitoring]
prometheus_monitoring_port = XXXX
```

Chainhook will behave as usual with this flag ommitted - metrics can still be retrieved via the observer's `/ping` endpoint, but they will not be formatted for ingestion by a Prometheus client.

---

### Checklist

- [X] All tests pass
- [X] Tests added in this PR (if applicable)

Fixes #474, addresses #466

### DIFF
--- a/components/chainhook-cli/src/config/mod.rs
+++ b/components/chainhook-cli/src/config/mod.rs
@@ -162,7 +162,7 @@ impl Config {
             http_api: match config_file.http_api {
                 None => PredicatesApi::Off,
                 Some(http_api) => match http_api.disabled {
-                    Some(false) => PredicatesApi::Off,
+                    Some(true) => PredicatesApi::Off,
                     _ => PredicatesApi::On(PredicatesApiConfig {
                         http_port: http_api.http_port.unwrap_or(DEFAULT_CONTROL_PORT),
                         display_logs: http_api.display_logs.unwrap_or(true),
--- a/components/chainhook-cli/src/config/mod.rs
+++ b/components/chainhook-cli/src/config/mod.rs
@@ -31,6 +31,7 @@ pub struct Config {
     pub event_sources: Vec<EventSourceConfig>,
     pub limits: LimitsConfig,
     pub network: IndexerConfig,
+    pub monitoring: MonitoringConfig,
 }
 
 #[derive(Clone, Debug, PartialEq)]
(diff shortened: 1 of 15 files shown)


components/chainhook-sdk/src/observer/mod.rs
```diff
@@ -136,59 +137,6 @@ impl EventObserverConfig {
             _ => unreachable!(),
         }
     }
-
-    pub fn new_using_overrides(
```
MicaiahReid, line 140: @lgalabru Just want to verify that this isn't being used by ordhook or some other dep and is safe to delete.

components/chainhook-cli/src/config/mod.rs
```diff
@@ -144,15 +150,19 @@ impl Config {
...
+        };
         let config = Config {
             storage: StorageConfig {
                 working_dir: config_file.storage.working_dir.unwrap_or("cache".into()),
             },
             http_api: match config_file.http_api {
                 None => PredicatesApi::Off,
                 Some(http_api) => match http_api.disabled {
-                    Some(false) => PredicatesApi::Off,
+                    Some(true) => PredicatesApi::Off,
```
lgalabru, line 165: 👏

components/chainhook-sdk/src/observer/mod.rs
lgalabru: Yeah this helper is indeed useful in the context of the sdk. Could you please add a comment?
It's being used in Pyth relayer - convenient way to DRY the code

https://github.com/hirosystems/stacks-pyth-bridge/blob/3afb7a8b174d3cc1a3caeac0890af4379a2e4879/example/relayer/src/config/mod.rs#L79

MicaiahReid: Added back with comment 👍

PR #480, opened 2024-01-25
fix: reduce memory usage in archive file ingestion 

MicaiahReid:
Previously, when parsing and archive tsv to determine the canonical fork, we would always start at block 0, even if there was a rocksdb instance containing stacks chaindata. The entire fork was loaded in memory and inserted.

Now, we start at the latest confirmed chaintip for this process, which greatly reduces the memory footprint on subsequent startups of the Chainhook service. Note: an initial run of the Chainhook service can still use significant memory.

### DIFF
--- a/components/chainhook-cli/src/scan/stacks.rs
+++ b/components/chainhook-cli/src/scan/stacks.rs
@@ -62,20 +62,20 @@ pub enum RecordKind {
 
 pub async fn get_canonical_fork_from_tsv(
     config: &mut Config,
+    start_block: Option<u64>,
     ctx: &Context,
 ) -> Result<VecDeque<(BlockIdentifier, BlockIdentifier, String)>, String> {
     let seed_tsv_path = config.expected_local_stacks_tsv_file().clone();
 
     let (record_tx, record_rx) = std::sync::mpsc::channel();
 
-    let start_block = 0;
-    let ctx_moved = ctx.clone();
+    let start_block = start_block.unwrap_or(0);
+    info!(
+        ctx.expect_logger(),
+        "Parsing tsv file to determine canoncial fork; starting at block {}", start_block
+    );
     let parsing_handle = hiro_system_kit::thread_named("Stacks chainstate CSV parsing")
         .spawn(move || {
-            info!(
-                ctx_moved.expect_logger(),
-                "Starting Stacks chainstate ingestion."
-            );
             let mut reader_builder = csv::ReaderBuilder::default()
                 .has_headers(false)
                 .delimiter(b'\t')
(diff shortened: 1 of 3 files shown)


components/chainhook-cli/src/storage/mod.rs
```diff
@@ -91,12 +91,15 @@ pub fn insert_entry_in_stacks_blocks(block: &StacksBlockData, stacks_db_rw: &DB,
...
         .put(&key, &block_bytes.to_string().as_bytes())
         .expect("unable to insert blocks");
-    stacks_db_rw
-        .put(
-            get_last_confirmed_insert_key(),
-            block.block_identifier.index.to_be_bytes(),
-        )
-        .expect("unable to insert metadata");
+    let previous_last_inserted = get_last_block_height_inserted(stacks_db_rw, _ctx).unwrap_or(0);
+    if block.block_identifier.index > previous_last_inserted {
```
MicaiahReid, line 95: Note that this makes `get_last_confirmed_insert_key` contain the _highest confirmed_ block, not necessarily the last inserted. This doesn't break tests, and I think it was the original intent of the function, but I figured it's worth noting.

components/chainhook-cli/src/scan/stacks.rs
```diff
@@ -95,6 +99,7 @@ pub async fn get_canonical_fork_from_tsv(
         })
         .expect("unable to spawn thread");
 
+    let stacks_db = open_readonly_stacks_db_conn(&config.expected_cache_path(), ctx).unwrap();
```
lgalabru, line 102: Instead of unwrapping, could we wait. loop and retry? opening could be temporarily impossible because another thread is writing, and that should be something that we're tolerating.

components/chainhook-cli/src/scan/stacks.rs
```diff
@@ -538,13 +556,19 @@ pub async fn consolidate_local_stacks_chainstate_using_csv(
 
     let _ = download_stacks_dataset_if_required(config, ctx).await;
 
-    let mut canonical_fork = get_canonical_fork_from_tsv(config, ctx).await?;
+    let stacks_db = open_readonly_stacks_db_conn(&config.expected_cache_path(), ctx).unwrap();
```
lgalabru, line 559: Same here. It probably worth introducing a new helper doing the loop / retry.

lgalabru (approved):
LGTM, thanks @MicaiahReid. Just a feedback on connection handling, feel free to create an issue for tracking if you think it's out of scope.

components/chainhook-cli/src/scan/stacks.rs
MicaiahReid: Done in commit 9cc2173b6c89b52b13fac9870dc1383ca09d8972

PR #486, opened 2024-02-01
fix: use cli feature for `cargo chainhook-install`

MicaiahReid:
Previously, the `cargo chainhook-install` script was installing the sdk version of Chainhook, which logged data in a non-user-friendly way. This changes our features some and updates the install script to fix those logs.


lgalabru (approved):
LGTM

PR #489, opened 2024-02-06
fix: validate predicate `start_block` and `end_block`

### LINKED ISSUE #464: Support future `start_block`
When `start_block` is set to a future block, the node is erroring, trying to reverse and scan the interval `(chain_tip...start_block)`.

### LINKED ISSUE #477: Chainhook should not error when end_block is missing; should just abort scan
[Here](https://github.com/hirosystems/chainhook/blob/abcf8096bd92ade907e0650371f08f475347843d/components/chainhook-cli/src/scan/stacks.rs#L192) and [here](https://github.com/hirosystems/chainhook/blob/abcf8096bd92ade907e0650371f08f475347843d/components/chainhook-cli/src/scan/stacks.rs#L202-L205), we throw an error if the end_block is not provided (or if we can't find blocks in the database).

I believe we should just revert to streaming mode in this case.

MicaiahReid:
### Description

Previously, the scanning threads had some validation for the `start_block` and `end_block` that was incorrect. This PR introduces validation that does the following:
 - We now allow `start_block` to be omitted by the user and we default to 0
 - If there are no blocks in the database, we abort the scan and go to streaming mode rather than erroring (fixes #477)
 - If the user provides an `end_block`, we validate that it is greater than the `start_block`
 - If the `start_block` is greater than chain tip, we abort the scan and go to streaming mode rather than erroring (fixes #464)

This PR also adds some validation to the `BlockHeights` class. Previously, it was possible to overload the `BlockHeights::BlockRange(start_block, end_block)` function to allocate a lot of memory into an empty array. We now have limits on this class. However, due to the above validation, it should not be possible to pass through parameters that reach theses limits (with the current usage of the class) until a chain height is up to `1_000_000`.

---

### Checklist

- [x] All tests pass

### DIFF
--- a/components/chainhook-cli/src/scan/bitcoin.rs
+++ b/components/chainhook-cli/src/scan/bitcoin.rs
@@ -111,13 +111,13 @@ pub async fn scan_bitcoin_chainstate_via_rpc_using_predicate(
                     ));
                 }
             };
-            // if not, break out so we can enter streaming mode
+            // if the chain hasn't progressed, break out so we can enter streaming mode
             // and put back the block we weren't able to scan
             if current_block_height > chain_tip {
                 block_heights_to_scan.push_front(current_block_height);
                 break;
             } else {
-                // if so, update our total number of blocks to scan
+                // if the chain has progressed, update our total number of blocks to scan and keep scanning
                 number_of_blocks_to_scan += chain_tip - prev_chain_tip;
             }
         }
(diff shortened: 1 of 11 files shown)


components/chainhook-cli/src/scan/bitcoin.rs
```diff
@@ -42,43 +42,70 @@ pub async fn scan_bitcoin_chainstate_via_rpc_using_predicate(
...
-                        (end_block, false)
-                    }
+        match BlockHeights::Blocks(blocks.clone()).get_sorted_entries() {
+            Ok(heights) => heights,
+            Err(e) => match e {
+                BlockHeightsError::ExceedsMaxEntries(max, specified) => {
+                    return Err(format!("Chainhook specification exceeds max number of blocks to scan. Maximum: {}, Attempted: {}", max, specified));
                 }
-                None => (result.blocks, true),
+                BlockHeightsError::StartLargerThanEnd => unreachable!(),
```
lgalabru, line 51: Could we return an error instead of panicking?

lgalabru (approved):
LGTM, thanks @MicaiahReid!

lgalabru (approved):
Thanks @MicaiahReid !

PR #490, opened 2024-02-07
fix: remove unreachable panic; return instead

MicaiahReid:
This error caused chainhook to crash today. This route can happen if a predicate is completed, but we're still waiting on block confirmations for fully expire the predicate, and then the user deletes the predicate.


lgalabru (approved):
LGTM!

PR #503, opened 2024-02-15
fix: log errors on block download failure; implement max retries

MicaiahReid:
### Description

Before this PR, Chainhook would loop infinitely if there is a problem downloading a BTC block, and it wouldn't log much information as to what was causing the failure. This PR implements better logging, and breaks out of the loop at 10 failed attempts.


csgui (approved):
Code LGTM!

Just a remark: maybe that solution can be improved as a configuration on the `Chainhook.toml` file at the `limits` table setings. Or even more granular, for instance, as a configuration in a predicate `json` file.

MicaiahReid:
@csgui Great call!

There's a lot to be desired still in our implementation of limits like these, I've made this issue to track! https://github.com/hirosystems/chainhook/issues/533

PR #505, opened 2024-02-15
fix: seed forking handler with unconfirmed blocks to improve startup stability

### LINKED ISSUE #487: store blocks from non-canonical forks; supply them to indexer on startup
If Chainhook restarts during a reorg, it needs to have some recent data to determine if incoming blocks are part of the canonical chain or not. This requires three changes:
- when starting up the event observer, the indexer's initial forks should be seeded with data (all unconfirmed blocks) from rocksdb
- we should store blocks from non-canonical forks in rocksdb, so the indexer has a full picture of the pre-shutdown state
- when a block is confirmed, it should prune the now invalid forks that were stored in rocksdb (to prevent a bloated db)

MicaiahReid:
### Description

If Chainhook restarts in the middle of a reorg taking place, it doesn't have any context for choosing the canonical fork. Then when block collisions take place, Chainhook fails to process the new blocks, causing gaps in the blocks Chainhook has available for evaluation.

This PR seeds the stacks block indexer with unconfirmed blocks on startup, so that Chainhook has the necessary context to handle a reorg.

Most of the PR is to add two tests:
 - I've added some new functionality to our very thorough indexer tests
   - In addition to providing the blocks to be mined and the order to mine them, we also now allow providing some "unconfirmed" blocks to seed the block pool with.
   - I've added some test cases that reproduce what caused outages on the Platform's Chainhook node
 - I've added a new service test that:
   - Verifies that unconfirmed blocks are stored on restart
   - Verifies that those blocks are used to seed the block pool, and that a reorg is handled correctly.

I committed these tests _before_ adding the fix, so you can confirm the fix by checking out the commits with the new tests (96d8249b239e53a877a98ff493ef7ae3571aca37 and 9aad55e2a88bf5b75c3e49d079c5967b7a8cf0e3), seeing that the tests break, then pulling the last commit to see that the fix works.

Fixes #487

### DIFF
--- a/components/chainhook-cli/src/service/mod.rs
+++ b/components/chainhook-cli/src/service/mod.rs
@@ -6,7 +6,8 @@ use crate::scan::stacks::consolidate_local_stacks_chainstate_using_csv;
 use crate::service::http_api::{load_predicates_from_redis, start_predicate_api_server};
 use crate::service::runloops::{start_bitcoin_scan_runloop, start_stacks_scan_runloop};
 use crate::storage::{
-    confirm_entries_in_stacks_blocks, draft_entries_in_stacks_blocks, open_readwrite_stacks_db_conn,
+    confirm_entries_in_stacks_blocks, draft_entries_in_stacks_blocks, get_all_unconfirmed_blocks,
+    open_readonly_stacks_db_conn_with_retry, open_readwrite_stacks_db_conn,
 };
 
 use chainhook_sdk::chainhooks::types::{ChainhookConfig, ChainhookFullSpecification};
@@ -20,7 +21,7 @@ use chainhook_sdk::types::{Chain, StacksChainEvent};
 use chainhook_sdk::utils::Context;
 use redis::{Commands, Connection};
 
-use std::sync::mpsc::channel;
+use std::sync::mpsc::{channel, Receiver, Sender};
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use self::http_api::get_entry_from_predicates_db;
@@ -38,6 +39,7 @@ impl Service {
     pub async fn run(
         &mut self,
         predicates_from_startup: Vec<ChainhookFullSpecification>,
+        observer_commands_tx_rx: Option<(Sender<ObserverCommand>, Receiver<ObserverCommand>)>,
     ) -> Result<(), String> {
         let mut chainhook_config = ChainhookConfig::new();
 
@@ -149,7 +151,8 @@ impl Service {
             }
         }
 
-        let (observer_command_tx, observer_command_rx) = channel();
+        let (observer_command_tx, observer_command_rx) =
+            observer_commands_tx_rx.unwrap_or(channel());
         let (observer_event_tx, observer_event_rx) = crossbeam_channel::unbounded();
         // let (ordinal_indexer_command_tx, ordinal_indexer_command_rx) = channel();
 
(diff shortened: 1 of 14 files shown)


components/chainhook-cli/src/service/mod.rs
```diff
@@ -38,6 +39,7 @@ impl Service {
     pub async fn run(
         &mut self,
         predicates_from_startup: Vec<ChainhookFullSpecification>,
+        observer_commands_tx_rx: Option<(Sender<ObserverCommand>, Receiver<ObserverCommand>)>,
```
MicaiahReid, line 42: Being able to create the observer commands tx externally allows stopping the service from a test.

components/chainhook-cli/src/service/tests/helpers/mock_service.rs
```diff
@@ -1,33 +1,37 @@
-use crate::config::Config;
```
MicaiahReid, line 1: This file is mostly just moving some test setup functions here from a different file, and adding the `TestSetupResult` struct to clean up the response a bit.

components/chainhook-cli/src/service/tests/helpers/mock_stacks_node.rs
```diff
@@ -8,7 +8,7 @@ use chainhook_sdk::types::{
     STXTransferEventData, SmartContractEventData, StacksTransactionEventPayload,
```
MicaiahReid, line 8: This file's changes are to allow adding a `fork_id` and `parent_fork_id` to the stacks blocks that are mined for tests.

components/chainhook-cli/src/service/tests/mod.rs
```diff
@@ -1,7 +1,5 @@
-use chainhook_sdk::chainhooks::types::{
-    ChainhookFullSpecification, ChainhookSpecification, StacksChainhookFullSpecification,
-};
-use chainhook_sdk::types::{Chain, StacksNetwork};
+use chainhook_sdk::chainhooks::types::ChainhookFullSpecification;
```
MicaiahReid, line 1: A lot of the changes in this file are more cleanup:
 - moving out helpers
 - our `mine_block` function now takes fork ids, so updating existing calls to that function to pass an fork id
 - our helpers have a new struct for the return value, so updating the tests that use that helper
 - simplifying our code to clean up temp test data

The actual change is the addition of the `it_saves_unconfirmed_blocks` test all the way at the bottom

components/chainhook-sdk/src/indexer/stacks/tests.rs
```diff
@@ -15,262 +15,271 @@ use test_case::test_case;
 
 #[test]
 fn test_stacks_vector_001() {
-    process_stacks_blocks_and_check_expectations(helpers::stacks_shapes::get_vector_001());
+    process_stacks_blocks_and_check_expectations((helpers::stacks_shapes::get_vector_001(), None));
```
MicaiahReid, line 18: I know it's not ideal that we have this additional argument being passed. The other option is to duplicate the `process_stacks_blocks_and_check_expectations` function with the changes I made to seed the block pool, so only the two new tests need to call it. Pick your poison ¯\_(ツ)_/¯

components/chainhook-sdk/src/indexer/tests/helpers/stacks_shapes.rs
```diff
@@ -3739,3 +3739,91 @@ pub fn get_vector_052() -> Vec<(BlockEvent, StacksChainEventExpectation)> {
         ),
     ]
 }
+
+/// Vector 053: Generate the following blocks
+///
+///  A1(0) - B1(1) - C1(3)
```
MicaiahReid, line 3745: > A1(0)
I've added the `(0)` to mean that this data already existed on startup (so we're seeding the block pool with this block)

Issue #498, opened 2024-02-13
[service] Revisit log levels


MicaiahReid:
**Describe the bug**
The chainhook service has a lot of `warn` and `error` logs that probably should be `debug` or `info`, including but not limited to:
 - The `"unable to load chainhook associated with key {}"` error when retrieving deleted predicates
 - `/new_mempool_tx`/`drop_mempool_tx`/unimplemented observer endpoints should be `debug` 
 - Possibly predicate trigger failing?
 - `Attempting to set _ status when previous status was _`

PR #511, opened 2024-03-01
fix: update scan status for non-triggering predicates

MicaiahReid:
### Description

Previously, we would only update the scanning status every 10 blocks _if_ the block we just scanned had a trigger. This leads to users not getting status updates on the predicates that don't trigger often.

Now, we update status every 10 blocks that we scan. This could lead to some noisy logs, so it should possibly be shipped with #498


lgalabru (approved):
LGTM!

PR #513, opened 2024-03-01
fix: skip db consolidation if no new dataset was downloaded

MicaiahReid:
### Description

The database consolidation loop doesn't need to happen if there is no new data. 

@lgalabru I'm interested in your thoughts on this. Potentially someone could break their stacks.rocksdb, then Chainhook wouldn't automatically fix it unless they deleted their archive to force a redownload, or a new archive becomes available.

I think the tradeoff is worth it because of the improvements to startup time if users are repeatedly running `chainhook predicates scan`, or if the service goes down and needs to restart.

### DIFF
--- a/components/chainhook-cli/src/scan/stacks.rs
+++ b/components/chainhook-cli/src/scan/stacks.rs
@@ -516,59 +516,66 @@ pub async fn consolidate_local_stacks_chainstate_using_csv(
         "Building local chainstate from Stacks archive file"
     );
 
-    let _ = download_stacks_dataset_if_required(config, ctx).await;
+    let downloaded_new_dataset = download_stacks_dataset_if_required(config, ctx).await;
 
-    let stacks_db = open_readonly_stacks_db_conn_with_retry(&config.expected_cache_path(), 3, ctx)?;
-    let confirmed_tip = get_last_block_height_inserted(&stacks_db, &ctx);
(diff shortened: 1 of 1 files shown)


components/chainhook-cli/src/scan/stacks.rs
```diff
@@ -516,59 +516,66 @@ pub async fn consolidate_local_stacks_chainstate_using_csv(
...
+        let mut canonical_fork = get_canonical_fork_from_tsv(config, confirmed_tip, ctx).await?;
+
+        let mut indexer = Indexer::new(config.network.clone());
+        let mut blocks_inserted = 0;
+        let mut blocks_read = 0;
+        let blocks_to_insert = canonical_fork.len();
+        let stacks_db_rw = open_readwrite_stacks_db_conn(&config.expected_cache_path(), ctx)?;
+        info!(
+            ctx.expect_logger(),
+            "Begining import of {} Stacks blocks into rocks db", blocks_to_insert
```
csgui, line 534:
```suggestion
            "Beginning import of {} Stacks blocks into rocks db", blocks_to_insert
```

csgui (approved):
Code LGTM.

About the mentioned tradeoff: I think that if the database is broken for any reason, is not Chainhook responsibility to fix it. But it should warn the issue somehow.

Thus, to improve the startup time - a thing that is inherent to Chainhook - my opinion is that the tradeoff of not try to fix the database is worth.

PR #514, opened 2024-03-01
feat: use stacks.rocksdb for predicate scan

### LINKED ISSUE #485: `chainhook predicates scan` should use rocksdb if present
https://github.com/hirosystems/chainhook/blob/5503c3d64d9f116bfd4cf71dfbe365b74219ec8d/components/chainhook-cli/src/cli/mod.rs#L481-L482

MicaiahReid:
### Description

Previously, when using the `chainhook predicates scan` command, Chainhook would download a very large archive file to use as a data source for the scan **every time the command was run**.

The Chainhook service (`chainhook service start`), however, implements rocksdb to allow storing this dataset locally. 

This PR uses the rocksdb that a user may have created locally to improve performance of the `chainhook predicates scan` command. If a rocksdb doesn't exist or can't be created, Chainhook will fallback to the original method of downloading the archive.

Combined with #513, this PR reduces a simple scan of a few blocks from ~20 minutes to .02 seconds.

Fixes #485

### DIFF
--- a/components/chainhook-cli/src/cli/mod.rs
+++ b/components/chainhook-cli/src/cli/mod.rs
@@ -3,6 +3,7 @@ use crate::config::Config;
 use crate::scan::bitcoin::scan_bitcoin_chainstate_via_rpc_using_predicate;
 use crate::scan::stacks::{
     consolidate_local_stacks_chainstate_using_csv, scan_stacks_chainstate_via_csv_using_predicate,
+    scan_stacks_chainstate_via_rocksdb_using_predicate,
 };
 use crate::service::http_api::document_predicate_api_server;
 use crate::service::Service;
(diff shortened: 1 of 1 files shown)


diwakergupta:
> Combined with #513, this PR reduces a simple scan of a few blocks from ~20 minutes to .02 seconds.

Whoa nicely done @MicaiahReid 👏🏽

components/chainhook-cli/src/cli/mod.rs
```diff
@@ -485,14 +486,35 @@ async fn handle_command(opts: Opts, ctx: Context) -> Result<(), String> {
...
+                                    &config,
+                                    &ctx,
+                                )
+                                .await?;
+                            }
+                            Err(e) => {
+                                warn!(
+                                    ctx.expect_logger(),
+                                    "Could not open db. This will greatly increase scan times. Error: {}", e
+                                );
```
csgui, line 509: I would expect a warning message to be something with a potential hazard on it. :-) 

Can that part `This will greatly increase scan times.` be on an `info!` message?

csgui:
@MicaiahReid PR looks good. But just added a comment of a thing that was bugging me. Waiting your feedback on that before finishing the review of this PR. Thanks!

components/chainhook-cli/src/cli/mod.rs
MicaiahReid: Done, good call!

csgui (approved):
LGTM! Great work Micaiah!

PR #515, opened 2024-03-04
fix(metrics): update latest ingested block on reorg

MicaiahReid:
Chainhook's `chainhook_stx_highest_block_ingested` and `chainhook_btc_highest_block_ingested` metrics were not being updated on a reorg. This was causing the Chainhook node to appear to be lagging behing the source chains, event though Chainhook's reorg handler was functioning properly.

This PR sets the highest block ingested metrics on a reorg based on the highest index block being applied in the reorg.


lgalabru (approved):
Thanks @MicaiahReid!

PR #542, opened 2024-03-28
chore: use stacks-codec

hugocaillard:
### Description

- replace the stacks-rpc-client dependency in favor of stacks-codec one
- replace `stacks_rpc_client::PoxInfo` with a new `PoxConfig` struct
  - I think that PoxInfo was misused here, it's supposed to represent the live state of pox on the chain, while chainhook-sdk is just storing some default config values


MicaiahReid (approved):
🎉

PR #543, opened 2024-03-29
fix: limit scope of stacks_rw_db usage to free lock

MicaiahReid:
Every 32 stacks blocks, Chainhook attempts to check for a new archive file to fix any potential holes in the database. This process requires readwrite access to the stacks.rocksdb. However, the process that called this subprocess already had a readwrite instance of the database open, which was preventing the lock from being cleared.

This PR sets the scope of the initial usage of the db in order to free up the lock.

### DIFF
--- a/components/chainhook-cli/src/service/mod.rs
+++ b/components/chainhook-cli/src/service/mod.rs
@@ -476,7 +476,7 @@ impl Service {
                                 ) {
                                     error!(
                                         self.ctx.expect_logger(),
-                                        "unable add confirmed entries to stacks db: {}", e
+                                        "unable to add confirmed entries to stacks db: {}", e
                                     );
                                 };
                                 if let Err(e) = draft_entries_in_stacks_blocks(
@@ -486,7 +486,7 @@ impl Service {
                                 ) {
                                     error!(
                                         self.ctx.expect_logger(),
-                                        "unable add unconfirmed entries to stacks db: {}", e
+                                        "unable to add unconfirmed entries to stacks db: {}", e
                                     );
                                 };
                             }
@@ -498,7 +498,7 @@ impl Service {
                                 ) {
                                     error!(
                                         self.ctx.expect_logger(),
-                                        "unable add confirmed entries to stacks db: {}", e
+                                        "unable to add confirmed entries to stacks db: {}", e
                                     );
                                 };
                                 if let Err(e) = draft_entries_in_stacks_blocks(
@@ -508,7 +508,7 @@ impl Service {
                                 ) {
                                     error!(
                                         self.ctx.expect_logger(),
-                                        "unable add unconfirmed entries to stacks db: {}", e
+                                        "unable to add unconfirmed entries to stacks db: {}", e
                                     );
                                 };
                             }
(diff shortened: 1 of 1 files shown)


components/chainhook-cli/src/service/mod.rs
```diff
@@ -463,11 +463,58 @@ impl Service {
...
+                        Ok(stacks_db_conn_rw) => match &chain_event {
+                            StacksChainEvent::ChainUpdatedWithBlocks(data) => {
+                                if let Err(e) = confirm_entries_in_stacks_blocks(
+                                    &data.confirmed_blocks,
+                                    &stacks_db_conn_rw,
+                                    &self.ctx,
+                                ) {
+                                    error!(
+                                        self.ctx.expect_logger(),
+                                        "unable add confirmed entries to stacks db: {}", e
```
csgui, line 479:
```suggestion
                                        "unable to add confirmed entries to stacks db: {}", e
```

components/chainhook-cli/src/service/mod.rs
```diff
@@ -463,11 +463,58 @@ impl Service {
...
+                                    );
+                                };
+                                if let Err(e) = draft_entries_in_stacks_blocks(
+                                    &data.new_blocks,
+                                    &stacks_db_conn_rw,
+                                    &self.ctx,
+                                ) {
+                                    error!(
+                                        self.ctx.expect_logger(),
+                                        "unable add unconfirmed entries to stacks db: {}", e
```
csgui, line 489: Same change for the phrase.

components/chainhook-cli/src/service/mod.rs
```diff
@@ -463,11 +463,58 @@ impl Service {
...
+                            }
+                            StacksChainEvent::ChainUpdatedWithReorg(data) => {
+                                if let Err(e) = confirm_entries_in_stacks_blocks(
+                                    &data.confirmed_blocks,
+                                    &stacks_db_conn_rw,
+                                    &self.ctx,
+                                ) {
+                                    error!(
+                                        self.ctx.expect_logger(),
+                                        "unable add confirmed entries to stacks db: {}", e
```
csgui, line 501: Same change for the phrase.

components/chainhook-cli/src/service/mod.rs
```diff
@@ -463,11 +463,58 @@ impl Service {
...
+                                    );
+                                };
+                                if let Err(e) = draft_entries_in_stacks_blocks(
+                                    &data.blocks_to_apply,
+                                    &stacks_db_conn_rw,
+                                    &self.ctx,
+                                ) {
+                                    error!(
+                                        self.ctx.expect_logger(),
+                                        "unable add unconfirmed entries to stacks db: {}", e
```
csgui, line 511: Same change for the phrase.

csgui (changes requested):
At first, I thought the code could be simplified. However, upon closer inspection, I noticed that the `data` struct differs between `StacksChainEvent::ChainUpdatedWithBlocks` and `StacksChainEvent::ChainUpdatedWithReorg`. As a result, a more effective code reorganization is possible, but it would require deeper changes that are beyond the scope of this pull request.

Code LGTM, just a small change on the error message.

csgui (approved):

PR #544, opened 2024-03-29
feat: update pox default settings

hugocaillard:
### Description

- update pox config default setting to use 20 / 5 on devnet
- update dependencies from the clarinet repo

Note: I realized that we had an other PoxConfig struct somewhere else in the repo 👎 
Something I'd like to fix in a future PR

### DIFF
--- a/components/chainhook-sdk/Cargo.toml
+++ b/components/chainhook-sdk/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "chainhook-sdk"
-version = "0.12.7"
+version = "0.12.8"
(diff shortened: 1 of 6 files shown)


MicaiahReid (approved):

MicaiahReid (approved):

MicaiahReid (approved):

components/chainhook-sdk/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "chainhook-sdk"
-version = "0.12.7"
+version = "0.12.8"
```
MicaiahReid, line 3:
```suggestion
version = "0.12.11"
```

PR #553, opened 2024-04-05
fix: revisit 7+ blocks reorg handling

MicaiahReid:


components/chainhook-cli/src/storage/mod.rs
```diff
@@ -148,12 +148,16 @@ pub fn insert_unconfirmed_entry_in_stacks_blocks(
...
         .map_err(|e| format!("unable to insert blocks: {}", e))?;
-    stacks_db_rw
-        .put(
-            get_last_unconfirmed_insert_key(),
-            block.block_identifier.index.to_be_bytes(),
-        )
-        .map_err(|e| format!("unable to insert metadata: {}", e))?;
+    let previous_last_inserted =
+        get_last_unconfirmed_block_height_inserted(stacks_db_rw, _ctx).unwrap_or(0);
+    if block.block_identifier.index > previous_last_inserted {
```
MicaiahReid, line 153: @lgalabru This is the one part we need to think through. 

We use the `get_last_unconfirmed_block_height_inserted` to know the latest unconfirmed block, which is where we start for the fork handler seed data.

If our helper inserts an earlier block (304/305), and we have existing higher unconfirmed blocks (311), do we want the database to be seeded with 304/305, or 304-311?

lgalabru: 304-311?

MicaiahReid: That's what I thought, and that's what this if statement should allow 👍

lgalabru (approved):
Ship it!

PR #573, opened 2024-04-29
fix: shut down observer on bitcoin block download failure

### LINKED ISSUE #572: Missing 1 bitcoin block download causes subsequent blocks to fail
Sometimes when the Chainhook service receives a `new_burn_block`, it then fails to download the block.

I'm not entirely sure why this happens, as the same block will be downloaded (at pretty much the exact same time) by Ordhook.

But, when this does happen, because we have a gap in our fork pool, we fail to append all future bitcoin blocks. We should:
1. increase the number of retries when downloading a block
2. auto-restart chainhook if we fail to download a block

MicaiahReid:
Fixes #572

### DIFF
--- a/components/chainhook-sdk/src/observer/http.rs
+++ b/components/chainhook-sdk/src/observer/http.rs
@@ -85,7 +85,7 @@ pub async fn handle_new_bitcoin_block(
                 ctx.try_log(|logger| {
                     slog::crit!(logger, "Could not shut down event observer: {e}")
                 });
-                panic!("Could not shut down event observer: {e}")
+                std::process::exit(1)
             }
         }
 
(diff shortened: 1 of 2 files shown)


components/chainhook-sdk/src/observer/http.rs
```diff
@@ -67,12 +67,33 @@ pub async fn handle_new_bitcoin_block(
...
+        });
+        match background_job_tx.lock() {
+            Ok(tx) => {
+                let _ = tx.send(ObserverCommand::Terminate);
+            }
+            Err(e) => {
+                ctx.try_log(|logger| {
+                    slog::crit!(logger, "Could not shut down event observer: {e}")
+                });
+                panic!("Could not shut down event observer: {e}")
```
lgalabru, line 88: should we `std::process::exit(1)` instead, just to be 100% certain that we're terminating the process?

MicaiahReid: Done!

PR #574, opened 2024-05-01
fix: set `Interrupted` status for streaming predicates that fail

### LINKED ISSUE #523: streaming predicates never set interrupted status
If a predicate is in streaming mode and it fails (for example when being triggered and sending the payload to the user), we don't set the status to interrupted. This means platform users will never see the errors, and if the user's service is no longer receiving requests we'll keep this predicate around for ever.

MicaiahReid:
Fixes #523


lgalabru (approved):
Great addition, thanks @MicaiahReid !

PR #579, opened 2024-05-08
fix: use `stacks_node_rpc_url` from config

MicaiahReid:


components/chainhook-types-rs/src/rosetta.rs
```diff
@@ -8,6 +8,7 @@ use std::collections::HashSet;
 use std::fmt::Display;
 use std::hash::{Hash, Hasher};
 
+pub const DEFAULT_STACKS_NODE_RPC: &str = "http://localhost:20443";
```
csgui, line 11: Since this constant will be reused through the codebase, maybe a better place for its definition might be within the root file definition of the `chainhook-types-rs` library (`./lib.rs`). What you think?

MicaiahReid: I like it, thanks for the suggestion!

MicaiahReid:
@csgui Thanks for the review! I realized the original was targeting the wrong branch (I recently switched to using just `main` rather than `main`/`develop`), so I had to do a rebase. I've addressed your comments though 👍

csgui (approved):
Code LGTM! Thanks for the changes, Micaiah.

PR #595, opened 2024-05-15
fix: print to console for unknown cli commands

MicaiahReid:
Previously, these were output as `CRIT` errors, but they should just be printed to the console.

### DIFF
--- a/components/chainhook-cli/src/cli/mod.rs
+++ b/components/chainhook-cli/src/cli/mod.rs
@@ -23,7 +23,7 @@ use chainhook_sdk::chainhooks::types::{
 };
 use chainhook_sdk::types::{BitcoinNetwork, BlockIdentifier, StacksNetwork};
 use chainhook_sdk::utils::{BlockHeights, Context};
-use clap::{Parser, Subcommand};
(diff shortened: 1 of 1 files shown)


components/chainhook-cli/src/cli/mod.rs
```diff
@@ -23,7 +23,7 @@ use chainhook_sdk::chainhooks::types::{
 };
 use chainhook_sdk::types::{BitcoinNetwork, BlockIdentifier, StacksNetwork};
 use chainhook_sdk::utils::{BlockHeights, Context};
-use clap::{Parser, Subcommand};
+use clap::{ErrorKind, Parser, Subcommand};
```
csgui, line 26: Unused `ErrorKind` import. Can we remove it?

csgui (changes requested):

components/chainhook-cli/src/cli/mod.rs
MicaiahReid: Done, thanks!

csgui (approved):
LGTM!

PR #601, opened 2024-05-30
fix: allow aborting a predicate scan

MicaiahReid:
Previously, when Chainhook was run as a service and the runloop to scan stacks/bitcoin predicates was set, we had no way to abort that scan. If a predicate was set to scan 1m blocks, but the user discovered the predicate was wrong and needed to delete, the user could delete the predicate from the store, but the scan thread had already started and would run until completion.

This PR adds an abort signal to an ongoing scan so that when a predicate is deregistered, the scan is canceled.

### DIFF
--- a/components/chainhook-cli/src/scan/common.rs
+++ b/components/chainhook-cli/src/scan/common.rs
@@ -69,5 +69,5 @@ pub fn get_block_heights_to_scan(
 pub enum PredicateScanResult {
     ChainTipReached,
     Expired,
-    Derigistered,
+    Deregistered,
 }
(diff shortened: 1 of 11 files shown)


components/chainhook-cli/src/scan/common.rs
```diff
@@ -65,3 +65,9 @@ pub fn get_block_heights_to_scan(
     };
     Ok(block_heights_to_scan)
 }
+
+pub enum PredicateScanResult {
+    ChainTipReached,
+    Expired,
+    Derigistered,
```
rafaelcr, line 72:
```suggestion
    Deregistered,
```

rafaelcr (approved):
LGTM, just a small typo in an Enum

PR #603, opened 2024-06-06
chore: update predicate scanning status on every trigger

MicaiahReid:
The predicate status is used to report data to users and to pick back up where the service left off on a restart. This change updates the scanning predicate status to be updated on every successful trigger of a scanning predicate, to be sure we don't scan blocks twice on a restart (if the status was set to a previous block, we'd start on that block at startup)

### DIFF
--- a/components/chainhook-cli/src/scan/bitcoin.rs
+++ b/components/chainhook-cli/src/scan/bitcoin.rs
@@ -101,7 +101,7 @@ pub async fn scan_bitcoin_chainstate_via_rpc_using_predicate(
     let mut loop_did_trigger = false;
     while let Some(current_block_height) = block_heights_to_scan.pop_front() {
         if let Some(ref mut predicates_db_conn) = predicates_db_conn {
-            if number_of_blocks_scanned % 10 == 0 
+            if number_of_blocks_scanned % 100 == 0 
                 || number_of_blocks_scanned == 0
                 // if the last loop did trigger a predicate, update the status
                 || loop_did_trigger
--- a/components/chainhook-cli/src/scan/bitcoin.rs
+++ b/components/chainhook-cli/src/scan/bitcoin.rs
@@ -98,9 +98,14 @@ pub async fn scan_bitcoin_chainstate_via_rpc_using_predicate(
     let mut last_scanned_block_confirmations = 0;
     let http_client = build_http_client();
 
+    let mut loop_did_trigger = false;
     while let Some(current_block_height) = block_heights_to_scan.pop_front() {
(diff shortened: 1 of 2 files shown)


rafaelcr (approved):
LGTM

components/chainhook-cli/src/scan/bitcoin.rs
```diff
@@ -98,9 +98,14 @@ pub async fn scan_bitcoin_chainstate_via_rpc_using_predicate(
...
     let http_client = build_http_client();
 
+    let mut loop_did_trigger = false;
     while let Some(current_block_height) = block_heights_to_scan.pop_front() {
         if let Some(ref mut predicates_db_conn) = predicates_db_conn {
-            if number_of_blocks_scanned % 10 == 0 || number_of_blocks_scanned == 0 {
+            if number_of_blocks_scanned % 10 == 0 
+                || number_of_blocks_scanned == 0
+                // if the last loop did trigger a predicate, update the status
+                || loop_did_trigger
```
csgui, line 107: Can't this conditional be replaced by `number_of_times_triggered > 1`?

components/chainhook-cli/src/scan/bitcoin.rs
```diff
@@ -177,6 +183,7 @@ pub async fn scan_bitcoin_chainstate_via_rpc_using_predicate(
             Ok(actions) => {
                 if actions > 0 {
                     number_of_times_triggered += 1;
+                    loop_did_trigger = true
```
csgui, line 186: Related to my previous comment: aren't these variable interchangeable?

components/chainhook-cli/src/scan/bitcoin.rs
MicaiahReid: No, they are not interchangeable. `number_of_times_triggered ` is a running total of the number of triggers. `loop_did_trigger` identifies if the _last loop_ triggered.

The goal of this PR is that the status that is stored in redis always stays updated with the latest triggers. This will prevent refiring a trigger on restart. (side note, we can probably now increase that mod to `number_of_blocks_scanned % 100`. We don't need to update redis as often for non-triggered predicates if we're sure we always keep updated every time we trigger.

If we used `number_of_times_triggered > 1`, we'd always update the status after a single trigger has happened for a predicate, which would be a lot of calls to redis.

Does that makes sense? We can get on a call if this isn't clear

csgui: Alright, thanks for clearing that up, Micaiah.

csgui (approved):
Code LGTM!

PR #612, opened 2024-07-02
feat: PoxConfig improvements 

### LINKED ISSUE #547: Accept PoxConfig as a parameter and normalize the type
chainhook-sdk currently has two ways of handling PoX settings:
- [PoxConfig in types.rs](https://github.com/hirosystems/chainhook/blob/bd300d88291dbef69e3289ea7807f23787f91d04/components/chainhook-sdk/src/chainhooks/types.rs#L568-L573)
- [PoxConfig in indexer/mod.rs](https://github.com/hirosystems/chainhook/blob/bd300d88291dbef69e3289ea7807f23787f91d04/components/chainhook-sdk/src/indexer/mod.rs#L26-L30) (previously `PoxInfo`)

(My bad for not noticing that when I refactored the `PoxInfo` part)

They have different but similar properties. The most confusing one is
- `reward_phase_len` is the full cycle length (reward + prepare)
- `reward_phase_block_length` is only the reward phase (cycle length - prepare)
Note: This confusion exists in many places (stacks-core, api, etc)

- [ ] These 2 types could be normalized into one

Both structs have defaults, which I think can only make sense for mainnet.
On testnet(s) and devnet(s) it doesn't makes much sense to have defaults because the values are often customized in the stacks-node config. On clarinet, it would even be great if we could customize these settings (`first_burnchain_block_height` and cycle lengths). So I think the default values could be removed (except for mainnet eventually). Chainhook should instead accept this as a parameter

- [ ] Accept PoxConfig settings as a parameter

tippenein:
closes #547 

- [x] remove redundant PoxConfig and update uses to the chainhooks::types version
- [x] allow PoxConfig settings as parameters

### DIFF
--- a/components/chainhook-sdk/src/indexer/mod.rs
+++ b/components/chainhook-sdk/src/indexer/mod.rs
@@ -3,7 +3,7 @@ pub mod fork_scratch_pad;
 pub mod stacks;
 
 use crate::{
-    chainhooks::types::{PoxConfig, POX_CONFIG_DEVNET, POX_CONFIG_MAINNET, POX_CONFIG_TESTNET},
+    chainhooks::types::PoxConfig,
     utils::{AbstractBlock, Context},
 };
 
@@ -35,9 +35,9 @@ impl StacksChainContext {
         StacksChainContext {
             asset_class_map: HashMap::new(),
             pox_config: match network {
-                StacksNetwork::Mainnet => POX_CONFIG_MAINNET,
-                StacksNetwork::Testnet => POX_CONFIG_TESTNET,
-                _ => POX_CONFIG_DEVNET,
+                StacksNetwork::Mainnet => PoxConfig::mainnet_default(),
+                StacksNetwork::Testnet => PoxConfig::testnet_default(),
+                _ => PoxConfig::devnet_default(),
             },
         }
     }
--- a/components/chainhook-sdk/src/indexer/stacks/mod.rs
+++ b/components/chainhook-sdk/src/indexer/stacks/mod.rs
@@ -330,7 +330,7 @@ pub fn standardize_stacks_block(
     let pox_cycle_length: u64 = (chain_ctx.pox_config.get_pox_cycle_len()).into();
     let current_len = u64::saturating_sub(
         block.burn_block_height,
-        1 + (chain_ctx.pox_config.genesis_block_height as u64),
+        1 + (chain_ctx.pox_config.first_burnchain_block_height as u64),
     );
     let pox_cycle_id: u32 = (current_len / pox_cycle_length).try_into().unwrap_or(0);
     let mut events: HashMap<&String, Vec<&NewEvent>> = HashMap::new();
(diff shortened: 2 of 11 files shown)


components/chainhook-sdk/src/indexer/mod.rs
```diff
@@ -22,45 +25,6 @@ pub struct AssetClassCache {
...
-    pub prepare_phase_block_length: u32,
-    pub reward_phase_block_length: u32,
-}
-
-impl PoxConfig {
-    pub fn mainnet_default() -> PoxConfig {
-        PoxConfig {
-            first_burnchain_block_height: 666050,
-            prepare_phase_block_length: 100,
-            reward_phase_block_length: 2000,
```
MicaiahReid, line 37: @hugocaillard The `reward_phase_block_length` that we are removing here differs slightly from what we're keeping above. The same goes for devnet/testnet. Is this at all cause for concern? Do you know which numbers are correct?

hugocaillard: @MicaiahReid 
Yeah I think it a concern.

Ideally we should have `reward_length + prepare_length = cycle_length`
But, in some places in clarinet, stacks-core, and chainhook, we have a wrong wording for that in way that
`cycle_length = reward_length`, so `actual_reward_length = reward_length - prepare_length`

Soooo, yeah, it's kind of a mess, and just here in chainhook, we have the issue, with a prepare phase of 100, and a reward phase of 2000 or 2100. 

But the true reward phase on mainnet is 2000, so the cycle length is 2100.

If we have places were we use it, we probably want adjust the maths

MicaiahReid: @tippenein, do you think that's enough info to work with? Maybe we can all get on a call to make sure it's clear how to approach this

hugocaillard: And this is great: https://github.com/hirosystems/chainhook/blob/pox-config-refactor-547/components/chainhook-sdk/src/chainhooks/types.rs#L591-L593

components/chainhook-sdk/src/indexer/stacks/mod.rs
```diff
@@ -327,12 +327,10 @@ pub fn standardize_stacks_block(
...
     ctx: &Context,
 ) -> Result<StacksBlockData, String> {
-    let pox_cycle_length: u64 = (chain_ctx.pox_config.prepare_phase_block_length
-        + chain_ctx.pox_config.reward_phase_block_length)
-        .into();
+    let pox_cycle_length: u64 = (chain_ctx.pox_config.get_pox_cycle_len()).into();
     let current_len = u64::saturating_sub(
         block.burn_block_height,
-        1 + (chain_ctx.pox_config.first_burnchain_block_height as u64),
+        1 + (chain_ctx.pox_config.genesis_block_height as u64),
```
hugocaillard, line 333: If it's not too big of a change, the wording `first_burnchain_block_height` seems to be more broadly used.

And I think it's easier to understand what it means

hugocaillard (changes requested):

PR #616, opened 2024-07-10
feat: PoxConfig improvements

### LINKED ISSUE #547: Accept PoxConfig as a parameter and normalize the type

tippenein:
closes #547 

- [x] remove redundant PoxConfig and update uses to the chainhooks::types version
- [x] allow PoxConfig settings as parameters


hugocaillard:
CI is broken @tippenein

hugocaillard (approved):
LGTM! Thanks @tippenein

PR #634, opened 2024-08-07
fix: optimize memory for stacks tsv import into rocksdb

rafaelcr:
This PR changes the way chainhook imports a Stacks node TSV into rocksdb.

Before, it loaded the entire canonical chinstate (including the full block JSON messages) onto a `VecDeque` in memory and then drained that data into rocksdb. This was a very memory intensive process which crashed our dev pods ever time it ran.

Now, the process was changed to a `VecDeque` that only keeps the line numbers of the TSV where the block data exists, so it can later read blocks from the file 1 by 1 and insert them into rocksdb.


rafaelcr:
Thanks @tippenein . Yep, it does, it runs a bit faster I assume because it doesn't have to deal with the very large data structure that was in place before but perhaps that's only in the dev env I was using to test

PR #639, opened 2024-08-15
fix: validate ts client predicates before registering

rafaelcr:
Changes the TS client server code so it now checks if a predicate UUID already exists and is active in the target Chainhook node. If it is, it reuses it. If it isn't it removes it and registers again.

### DIFF
--- a/components/client/typescript/src/server.ts
+++ b/components/client/typescript/src/server.ts
@@ -11,7 +11,13 @@ import { request } from 'undici';
 import { logger, PINO_CONFIG } from './util/logger';
 import { timeout } from './util/helpers';
 import { Payload, PayloadSchema } from './schemas/payload';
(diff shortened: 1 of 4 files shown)


components/client/typescript/src/server.ts
```diff
@@ -118,16 +122,61 @@ export async function buildServer(
...
+            // Predicate doesn't exist.
+            break;
+          case true:
+            logger.info(
+              `ChainhookEventObserver predicate ${predicate.uuid} is already active, skipping registration`
+            );
+            continue;
+          case false:
+            logger.info(
+              `ChainhookEventObserver predicate ${predicate.uuid} was being used but is now inactive, updating`
```
csgui, line 174: This log information suggests that an inactive predicate will be updated somehow, rather than deleted. Perhaps changing the information to `... was being used but is now inactive, removing it` would be more accurate.

csgui (approved):
Code LGTM. Just a small comment on a log info.

PR #642, opened 2024-09-10
fix: add a configurable max timeout for outgoing predicate payload requests

### LINKED ISSUE #643: add a configurable max timeout for predicate requests
If chainhook is sending a POST request for a predicate into an external server but that server never responds, it leads to an indefinite lock in the thread that handles predicate evaluations, which disrupts the entire block consumption process.

We should add a configurable timeout for these requests.

rafaelcr:
This PR adds a new (optional) entry to the config .toml file so outgoing predicate payload HTTP POST requests can have a configurable timeout that will cancel the request if hit.

```toml
[predicates]
payload_http_request_timeout_ms = 10000
```

This new setting is useful to public-facing chainhook instances that may run into unresponsive 3rd party servers that may take forever to respond to a payload request.

Fixes #643


components/chainhook-cli/src/scan/stacks.rs
```diff
@@ -338,7 +340,12 @@ pub async fn scan_stacks_chainstate_via_rocksdb_using_predicate(
             apply: hits_per_blocks,
             rollback: vec![],
         };
-        let res = match handle_stacks_hook_action(trigger, &proofs, ctx) {
+        let res = match handle_stacks_hook_action(
+            trigger,
+            &proofs,
+            &config.get_event_observer_config().predicates_config,
```
csgui, line 346: It seems more reasonable to pass the parent `EventObserverConfig` here and let the `handle_stacks_hook_action` function decide what it needs. The caller don't need to know about implementation details of the callee function.

components/chainhook-cli/src/scan/bitcoin.rs
```diff
@@ -290,7 +290,7 @@ pub async fn execute_predicates_action<'a>(
             gather_proofs(&trigger, &mut proofs, config, ctx);
         }
         let predicate_uuid = &trigger.chainhook.uuid;
-        match handle_bitcoin_hook_action(trigger, &proofs) {
+        match handle_bitcoin_hook_action(trigger, &proofs, &config.predicates_config) {
```
csgui, line 293: It seems more reasonable to pass the parent `EventObserverConfig` here and let the `handle_bitcoin_hook_action` function decide what it needs. The caller don't need to know about implementation details of the callee function.

components/chainhook-cli/src/scan/stacks.rs
```diff
@@ -525,7 +534,12 @@ pub async fn scan_stacks_chainstate_via_csv_using_predicate(
             apply: hits_per_blocks,
             rollback: vec![],
         };
-        match handle_stacks_hook_action(trigger, &proofs, ctx) {
+        match handle_stacks_hook_action(
+            trigger,
+            &proofs,
+            &config.get_event_observer_config().predicates_config,
```
csgui, line 540: Same comment above.

components/chainhook-sdk/src/chainhooks/stacks/mod.rs
```diff
@@ -1325,11 +1327,16 @@ pub fn serialize_stacks_payload_to_json<'a>(
...
     match &trigger.chainhook.action {
         HookAction::HttpPost(http) => {
-            let client = Client::builder()
+            let mut client_builder = Client::builder();
+            if let Some(timeout) = config.payload_http_request_timeout_ms {
+                client_builder = client_builder.timeout(Duration::from_millis(timeout));
+            }
+            let client = client_builder
                 .build()
                 .map_err(|e| format!("unable to build http client: {}", e))?;
```
csgui, line 1341: What you think in exchange mutability for method chaining? A mutable `client` instance is less robust than an immutable one, IMO.

And with `unwrap_or_else` a default timeout can be set.

```rust
let client = Client::builder()
    .timeout(config.payload_http_request_timeout_ms
        .map(Duration::from_millis)
        .unwrap_or_else(|| Duration::from_secs(30))) // <<<--- Default timeout
    .build()
    .map_err(|e| format!("unable to build http client: {}", e))?;
```

csgui (changes requested):
Added some comments and changes as suggestion.

components/chainhook-sdk/src/chainhooks/stacks/mod.rs
rafaelcr: I agree with your suggestion, the problem is that we want the default to be "no timeout" and there's no way to set that via the `ClientBuilder`. I'm not sure why they don't have a method to provide an `Optional<Duration>`

csgui: Ok. 👍

csgui (approved):
Code LGTM!

PR #648, opened 2024-10-03
fix: return 500 status code to Stacks node when event ingestion fails

### LINKED ISSUE #646: return 500 status codes properly when a Stacks block could not be ingested
The HTTP server that consumes data from the Stacks node always returns a `200` status code regardless if the block has been ingested correctly or not, because the response handlers never actually return an error, the error is only reflected in the body:
```rust
return Json(json!({
    "status": 500,
    "result": format!("Unable to standardize stacks block {}", e),
}));
```
Because of this, the Stacks node continues to advance the chain creating gaps in the index that are impossible to recover from. This is related to #623

rafaelcr:
This PR makes sure we return a `500` status code error to the Stacks node when chainhook failed to ingest an event correctly, so it can retry indefinitely until it succeeds.

This prevents block gaps from occurring in the chainhook index.

Fixes #646


components/chainhook-sdk/src/observer/http.rs
```diff
@@ -58,116 +79,76 @@ pub async fn handle_new_bitcoin_block(
...
-                ctx.try_log(|logger| {
-                    slog::warn!(
-                        logger,
-                        "unable to download_and_parse_block: {}",
-                        e.to_string()
-                    )
-                });
-                None
+                return error_response(
+                    format!("unable to download_and_parse_block: {}", e.to_string()),
```
csgui, line 85: `e`is a type that implements `Display`. Change `e.to_string()` -> `e`.

components/chainhook-sdk/src/observer/http.rs
```diff
@@ -58,116 +79,76 @@ pub async fn handle_new_bitcoin_block(
...
-            return Json(json!({
-                "status": 500,
-                "result": "Unable to acquire lock",
-            }));
+            return error_response(
+                format!("unable to acquire background_job_tx: {}", e.to_string()),
+                ctx,
+            );
         }
     };
```
csgui, line 110: It seems that this pattern ignoring success cases can be improved. The empty case can be used as an opportunity to log success, for instance. Also is possible to reduce the nested matches.

Follow a suggestion (_not tested_) to reduce nesting and improve the code readability.

```rust
let bg_job_tx_mutex = background_job_tx.lock()
    .map(|tx| tx.send(ObserverCommand::ProcessBitcoinBlock(block))
    .map_err(|e| format!("Unable to send stacks chain event: {}", e)));

match bg_job_tx_mutex {
    Ok(_) => {
        try_info!(ctx, "Successfully sent ProcessBitcoinBlock command");
    }
    Err(e) => {
        let error_msg = format!("Unable to acquire bg_job_tx mutex: {}", e);
        try_error!(ctx, "{}", error_msg);
        return error_response(error_msg, ctx);
    }
}
```

components/chainhook-sdk/src/observer/http.rs
```diff
@@ -58,116 +79,76 @@ pub async fn handle_new_bitcoin_block(
...
-                    "unable to acquire indexer_rw_lock: {}",
-                    e.to_string()
-                )
-            });
-            return Json(json!({
-                "status": 500,
-                "result": "Unable to acquire lock",
-            }));
+            return error_response(
+                format!("Unable to acquire indexer_rw_lock: {}", e.to_string()),
```
csgui, line 116: `e.to_string()` -> `e`.

components/chainhook-sdk/src/observer/http.rs
```diff
@@ -58,116 +79,76 @@ pub async fn handle_new_bitcoin_block(
...
-                }
+                Ok(tx) => match tx.send(ObserverCommand::PropagateBitcoinChainEvent(chain_event)) {
+                    Ok(_) => {}
+                    Err(e) => {
+                        return error_response(
+                            format!("Unable to send stacks chain event: {}", e.to_string()),
+                            ctx,
+                        );
+                    }
+                },
```
csgui, line 134: Please check the comment above about ignoring success cases and nested statements.

components/chainhook-sdk/src/observer/http.rs
```diff
@@ -222,34 +196,32 @@ pub fn handle_new_stacks_block(
...
-                }
+                Ok(tx) => match tx.send(ObserverCommand::PropagateStacksChainEvent(chain_event)) {
+                    Ok(_) => {}
+                    Err(e) => {
+                        return error_response(
+                            format!("Unable to send stacks chain event: {}", e.to_string()),
+                            ctx,
+                        );
+                    }
+                },
```
csgui, line 207: Please check the comment above about ignoring success cases and nested statements.

components/chainhook-sdk/src/observer/http.rs
```diff
@@ -262,99 +234,102 @@ pub fn handle_new_microblocks(
...
```
csgui, line 264: Please check the comment above about ignoring success cases and nested statements.

components/chainhook-sdk/src/observer/http.rs
```diff
@@ -262,99 +234,102 @@ pub fn handle_new_microblocks(
...
+                }
+            }
+        }
+        Err(e) => {
+            return error_response(
+                format!("Unable to acquire background_job_tx: {}", e.to_string()),
+                ctx,
+            );
+        }
+    }
```
csgui, line 330: Please check the comment above about ignoring success cases and nested statements.

components/chainhook-sdk/src/observer/http.rs
```diff
@@ -203,17 +184,10 @@ pub fn handle_new_stacks_block(
...
-                    "unable to acquire indexer_rw_lock: {}",
-                    e.to_string()
-                )
-            });
-            return Json(json!({
-                "status": 500,
-                "result": "Unable to acquire lock",
-            }));
+            return error_response(
+                format!("Unable to acquire indexer_rw_lock: {}", e.to_string()),
```
csgui, line 188: `e.to_string()` -> `e`.

components/chainhook-sdk/src/observer/http.rs
```diff
@@ -222,34 +196,32 @@ pub fn handle_new_stacks_block(
...
-                            "unable to acquire background_job_tx: {}",
-                            e.to_string()
-                        )
-                    });
-                    return Json(json!({
-                        "status": 500,
-                        "result": "Unable to acquire lock",
-                    }));
+                    return error_response(
+                        format!("Unable to acquire background_job_tx: {}", e.to_string()),
```
csgui, line 210: `e.to_string()` -> `e`.

components/chainhook-sdk/src/observer/http.rs
```diff
@@ -262,99 +234,102 @@ pub fn handle_new_microblocks(
...
-                    "unable to acquire background_job_tx: {}",
-                    e.to_string()
-                )
-            });
-            return Json(json!({
-                "status": 500,
-                "result": "Unable to acquire lock",
-            }));
+            return error_response(
+                format!("Unable to acquire background_job_tx: {}", e.to_string()),
```
csgui, line 246: `e.to_string()` -> `e`.

components/chainhook-sdk/src/utils/mod.rs
```diff
@@ -403,3 +403,43 @@ pub fn write_file_content_at_path(file_path: &PathBuf, content: &[u8]) -> Result
...
+
+#[macro_export]
+macro_rules! try_error {
+    ($a:expr, $tag:expr, $($args:tt)*) => {
+        $a.try_log(|l| slog::error!(l, $tag, $($args)*));
+    };
+    ($a:expr, $tag:expr) => {
+        $a.try_log(|l| slog::error!(l, $tag));
+    };
+}
```
csgui, line 445: There’s a lot of code duplication here. Can we leverage a single core macro to try to reduce the duplication? 🤔 

I am thinking in some like this pseudo code below:

```
#[macro_export]
macro_rules! try_log {
    ... generic macro code here
}

#[macro_export] macro_rules! try_info  { (args) => { try_log!(info,  $(args) }; }
#[macro_export] macro_rules! try_debug { (args) => { try_log!(info,  $(args) }; }
#[macro_export] macro_rules! try_warn  { (args) => { try_log!(info,  $(args) }; }
#[macro_export] macro_rules! try_error { (args) => { try_log!(info,  $(args) }; }

```

csgui (changes requested):
Please check my comments. Maybe we can try to improve the implementation a bit. Thanks.

components/chainhook-sdk/src/observer/http.rs
rafaelcr: Thanks for the suggestion! I just cleaned these parts of the code. However, I don't think logging the success case is necessary because it would add some noise and not provide value to the logs. Let me know if you disagree though.

components/chainhook-sdk/src/utils/mod.rs
rafaelcr: Hmm I tried this actually but not sure why it doesn't let me refer to a macro defined in the same file from another macro

csgui: Let's consider this as a future improvement.

csgui (approved):
Code LGTM!

PR #653, opened 2024-10-14
feat: parse `/stackerdb_chunks` Stacks node event

rafaelcr:
Starts listening to the new `/stackerdb_chunks` Stacks node event and parses the incoming message. Parsers are taken from the `stacks_codec` crate, which has been updated with the latest serializers.

For future PRs:
* Start integrating the signer messages into the predicates framework
* Store signer messages in a local SQLite DB for replay
* Expand unit tests to include full Stacks node integration tests

### DIFF
--- a/components/chainhook-sdk/src/indexer/stacks/mod.rs
+++ b/components/chainhook-sdk/src/indexer/stacks/mod.rs
@@ -641,7 +641,7 @@ pub fn standardize_stacks_stackerdb_chunks(
             SignerMessage::BlockResponse(block_response) => match block_response {
                 BlockResponse::Accepted((block_hash, sig)) => StacksSignerMessage::BlockResponse(
                     BlockResponseData::Accepted(BlockAcceptedResponse {
-                        block_hash: block_hash.to_hex(),
+                        signer_signature_hash: block_hash.to_hex(),
                         sig: sig.to_hex(),
                     }),
                 ),
@@ -763,7 +763,8 @@ pub fn get_signer_pubkey_from_stackerdb_chunk_slot(
 
     let mut digest_bytes = slot.slot_id.to_be_bytes().to_vec();
     digest_bytes.extend(slot.version.to_be_bytes().to_vec());
-    digest_bytes.extend(data_bytes.clone());
+    let data_bytes_hashed = Sha512Trunc256Sum::from_data(&data_bytes).to_bytes();
+    digest_bytes.extend(data_bytes_hashed);
     let digest = Sha512Trunc256Sum::from_data(&digest_bytes).to_bytes();
 
     let sig_bytes =
--- a/components/chainhook-sdk/src/indexer/stacks/mod.rs
+++ b/components/chainhook-sdk/src/indexer/stacks/mod.rs
@@ -703,11 +703,9 @@ pub fn standardize_stacks_stackerdb_chunks(
         };
         parsed_chunks.push(StacksStackerDbChunk {
             contract: contract_id.clone(),
+            sig: slot.sig.clone(),
+            pubkey: get_signer_pubkey_from_stackerdb_chunk_slot(slot, &data_bytes)?,
             message,
-            raw_data: slot.data.clone(),
-            raw_sig: slot.sig.clone(),
-            slot_id: slot.slot_id,
-            slot_version: slot.version,
         });
     }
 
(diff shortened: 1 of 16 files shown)


components/chainhook-sdk/src/indexer/stacks/mod.rs
```diff
@@ -562,6 +591,198 @@ pub fn standardize_stacks_microblock_trail(
...
+        key::Secp256k1, secp256k1::{
+            ecdsa::{RecoverableSignature, RecoveryId},
+            Message,
+        }
+    };
+
+    let mut digest_bytes = slot.slot_id.to_be_bytes().to_vec();
+    digest_bytes.extend(slot.version.to_be_bytes().to_vec());
+    digest_bytes.extend(data_bytes.clone());
+    let digest = Sha512Trunc256Sum::from_data(&digest_bytes).to_bytes();
```
zone117x, line 764: `data_bytes` must be hashed first by itself, e.g. 
```ts
/** Get the digest to sign that authenticates this chunk data and metadata */
function authDigest(slot: ModifiedSlot): Buffer {
  const hasher = crypto.createHash('sha512-256');
  hasher.update(toU32BeBytes(slot.slot_id));
  hasher.update(toU32BeBytes(slot.slot_version));

  // Calculate the hash of the chunk bytes. This is the SHA512/256 hash of the data
  const dataHash = crypto.hash('sha512-256', Buffer.from(slot.data, 'hex'), 'buffer');
  hasher.update(dataHash);

  return hasher.digest();
}

```

components/chainhook-sdk/src/indexer/stacks/mod.rs
```diff
@@ -562,6 +591,198 @@ pub fn standardize_stacks_microblock_trail(
...
+        };
+        let signer_message =
+            match SignerMessage::consensus_deserialize(&mut Cursor::new(&data_bytes)) {
+                Ok(message) => message,
+                Err(e) => return Err(format!("unable to deserialize SignerMessage: {e}")),
+            };
+        let message = match signer_message {
+            SignerMessage::BlockProposal(block_proposal) => {
+                StacksSignerMessage::BlockProposal(BlockProposalData {
+                    block: standardize_stacks_nakamoto_block(&block_proposal.block),
```
zone117x, line 634: We should also include the `block_hash`, it can be determined by hashing the header fields:

(rust) https://github.com/stacks-network/stacks-core/blob/a2dcd4c3ffdb625a6478bb2c0b23836bc9c72f9f/stackslib/src/chainstate/nakamoto/mod.rs#L764-L795

(js) https://github.com/hirosystems/signer-monitor/blob/d67b153088ffad9ffcd9249301188d86b09dee95/chunk-parser/src/signer-message.ts#L184-L198

The `index_block_hash` can also be included easily:
```ts
function getNakamotoIndexBlockHash(blockHash: string, consensusHash: string): string {
  const hasher = crypto.createHash('sha512-256');
  hasher.update(Buffer.from(blockHash, 'hex'));
  hasher.update(Buffer.from(consensusHash, 'hex'));
  return hasher.digest('hex');
}
```

components/chainhook-sdk/src/indexer/stacks/mod.rs
```diff
@@ -562,6 +591,198 @@ pub fn standardize_stacks_microblock_trail(
...
+                    burn_height: block_proposal.burn_height,
+                    reward_cycle: block_proposal.reward_cycle,
+                })
+            }
+            SignerMessage::BlockResponse(block_response) => match block_response {
+                BlockResponse::Accepted((block_hash, sig)) => StacksSignerMessage::BlockResponse(
+                    BlockResponseData::Accepted(BlockAcceptedResponse {
+                        block_hash: block_hash.to_hex(),
```
zone117x, line 642: I actually had this mislabeled initially in the js code. Technically it's actually `signer_signature_hash` (just like in the `BlockResponse::Rejected` case), however, it happens to be the same value as `block_hash`. Probably best to use the `signer_signature_hash` naming in case it doesn't match block_hash in epoch3.x.

components/chainhook-sdk/src/observer/http.rs
```diff
@@ -178,6 +178,25 @@ pub fn handle_new_stacks_block(
...
+pub fn handle_stackerdb_chunks(
+    indexer_rw_lock: &State<Arc<RwLock<Indexer>>>,
+    payload: Json<JsonValue>,
+    ctx: &State<Context>,
+) -> Result<Json<JsonValue>, Custom<Json<JsonValue>>> {
+    try_info!(ctx, "POST /stackerdb_chunks");
+    // Standardize the structure of the StackerDB chunk, and identify the kind of update that this new message would imply.
+    let chain_event = match indexer_rw_lock.inner().write() {
+        Ok(mut indexer) => indexer
+            .handle_stacks_marshalled_microblock_trail(payload.into_inner(), ctx),
```
zone117x, line 192: Typo?
```suggestion
            .handle_stacks_marshalled_stackerdb_chunk(payload.into_inner(), ctx),
```

components/chainhook-sdk/src/indexer/stacks/mod.rs
zone117x: (can be done in another PR)

components/chainhook-sdk/src/observer/http.rs
```diff
@@ -178,6 +178,25 @@ pub fn handle_new_stacks_block(
     success_response()
 }
 
+#[cfg(feature = "stacks-signers")]
+#[post("/stackerdb_chunks", format = "application/json", data = "<payload>")]
+pub fn handle_stackerdb_chunks(
```
zone117x, line 183: Is this function the best time to grab `SystemTime::now()`? Or is the receipt time already handled elsewhere?

components/chainhook-sdk/src/observer/http.rs
rafaelcr: Whoops, thanks

components/chainhook-sdk/src/indexer/stacks/mod.rs
rafaelcr: Thanks, done

rafaelcr:
Thanks for the comments @zone117x I'll be adding some unit tests today with your test payloads so we can merge this one ASAP

rafaelcr:
@zone117x can you take a look? I added a unit test to check the parser behavior, but will expand the tests in a future PR. Right now, Chainhook has a strange way of depending on a local install of `redis` to run tests so it's tricky to run them in a local setup. I'll move that over to docker in the next PR and start adding some integration tests.

zone117x (approved):
LGTM!

PR #655, opened 2024-10-16
build: run redis tests using a docker integration environment

rafaelcr:
Changes how redis tests are performed so they run with a dockerized integration environment instead of relying on a local `redis-server` install controlled via Rust code.

Adds tasks and launch configs to VScode so they are easier to launch and adjusts the CI to use it as well.


zone117x (approved):
Tested the vscode launch configs locally and works great!

PR #656, opened 2024-10-18
feat: include signer messages in Stacks predicate payloads

rafaelcr:
* Add an `events` entry to Stacks predicate payloads that includes signer messages. To make it future proof, it takes in a struct that could include other events in the future.
* Update the TS client library to include the new payloads

This initial implementation evaluates a predicate for signer messages using only timestamps. In the future we'll expand this to add more functionality.


zone117x (commented):
So far this approach is looking good to me 👍

zone117x (approved):
Looks good!
Btw there was a last minute change to the codec in core, it now includes a metadata property on a few of these payloads, see https://github.com/hirosystems/signer-monitor/commit/ee7b0cedb23532f89533a1db53e01392376b1b4f

PR #657, opened 2024-10-18
feat(ts-client): add signer messages to stacks payloads

rafaelcr:
Expands the TS client to include new signer messages, and adds missing `signer_bitvec` and `signer_signatures` fields to block metadata.


rafaelcr:
@zone117x the easiest would be for you to checkout this branch locally and build/install the npm lib into your API repo so you can use it in your mock tests. Take a look at the [Metadata API](https://github.com/hirosystems/token-metadata-api/blob/develop/src/chainhook/server.ts) for an example on how to use it, also check out its unit tests which would give you an idea on how I usually write tests for payloads

zone117x (approved):

PR #658, opened 2024-10-22
feat(ts-client): add persisted predicates and periodic health check

rafaelcr:
Upgrades TS client to new breaking change with updated schema treatments and unit tests.

**Breaking changes**
* Instead of letting users define a predicate `uuid` by hand, it is now configured automatically by the client so users can only set the `name` and we can use that as a key to figure out if we had already registered a predicate before. This makes UUIDs be local to that running event observer and avoids collisions when pointing multiple observers to one single chainhook node.
* Upgrades schema names and configuration options to better reflect what they do

**Other features**
* Adds on-disk predicate persistence for registered predicates: saves to disk and recalls upon restarts
* Adds a configurable health check to make sure on-disk predicates are still healthy, and re-registers them if they are interrupted
* Adds unit tests for new behavior


components/client/typescript/src/index.ts
```diff
@@ -20,29 +102,41 @@ import {
...
+    callback: OnPredicatePayloadCallback
+  ): Promise<void> {
     if (this.fastify) return;
-    this.fastify = await buildServer(this.serverOpts, this.chainhookOpts, predicates, callback);
-    await this.fastify.listen({ host: this.serverOpts.hostname, port: this.serverOpts.port });
+    this.fastify = await buildServer(this.observer, this.chainhook, predicates, callback);
+    await this.fastify.listen({ host: this.observer.hostname, port: this.observer.port });
+    if (this.observer.predicate_health_check_interval_ms && this.healthCheckTimer === undefined) {
+      this.healthCheckTimer = setInterval(() => {
+        void predicateHealthCheck(this.observer, this.chainhook);
```
zone117x, line 129: Can `predicateHealthCheck` ever throw? If so, this needs a `.catch(err => log(err))` otherwise could end up with an unhandled promise rejection

zone117x (approved):
Looks robust 💯

Issue #651, opened 2024-10-08
add predicates for stacks signers


rafaelcr:
Chainhook should add code to parse StackerDB messages from the Stacks node and add predicates so listeners can get information on signer activity

PR #659, opened 2024-10-22
feat: add various new nakamoto block fields to `/new_block` ingestion and `StacksPayload`

zone117x:
Related to https://github.com/hirosystems/chainhook/issues/651 -- we need these new fields to make the `StackerDB` chunk messages useful.


components/chainhook-sdk/src/indexer/stacks/mod.rs
```diff
@@ -35,6 +35,36 @@ pub struct NewBlock {
...
+    pub signer_bitvec: Option<String>,
+    
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub signer_signature: Option<Vec<String>>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub cycle_number: Option<u64>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub reward_set: Option<RewardSet>,
```
zone117x, line 52: stacks-core isn't very consistent with some of these fields being optional (undefined) vs nullable. So goal here is to just support either case during deserialization.

components/chainhook-types-rs/src/rosetta.rs
```diff
@@ -114,6 +114,29 @@ pub struct StacksBlockMetadata {
...
+    pub pox_ustx_threshold: String,
+    pub rewarded_addresses: Vec<String>,
+    pub signers: Option<Vec<StacksBlockMetadataRewardSetSigner>>,
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+pub struct StacksBlockMetadataRewardSetSigner {
+    pub signing_key: String,
+    pub weight: u32,
+    pub stacked_amt: String,
```
zone117x, line 139: I'm simply re-using the same structure as the `/new_block` event for this payload so for now it looks like code duplication, however, it gives us room to restructure/rename/re-type these as needed.

components/chainhook-sdk/src/indexer/stacks/mod.rs
```diff
@@ -432,6 +462,29 @@ pub fn standardize_stacks_block(
             pox_cycle_length: pox_cycle_length.try_into().unwrap(),
             confirm_microblock_identifier,
             stacks_block_hash: block.block_hash.clone(),
+
+            block_time: block.block_time,
+            // TODO: decode `signer_bitvec` into an easy to use bit string representation (e.g. "01010101")
+            signer_bitvec: block.signer_bitvec.clone(),
```
zone117x, line 468: @rafaelcr we get these from `/new_block` in consesus-serialize format, e.g. `0x000800000001ff`. Is there existing code I can call to decode this? We should deliver it in the payloads in regular bitfield format (a string of `1` and/or `0`).

rafaelcr (approved):
Looks great! Thanks @zone117x

components/chainhook-sdk/src/indexer/stacks/mod.rs
rafaelcr: Yes, I believe so, it's probably included in the update we did to the `stacks-codec` here https://github.com/hirosystems/clarinet/pull/1524

PR #660, opened 2024-10-22
fix: increase stacks event server body size limit to 500MB

rafaelcr:


CharlieC3 (approved):
Thanks for the quick response! :shipit:

PR #662, opened 2024-10-24
feat: include recovered signer pubkeys in new block payload

zone117x:


components/chainhook-sdk/src/indexer/stacks/mod.rs
```diff
@@ -502,6 +512,19 @@ pub fn standardize_stacks_block(
...
+                    Some(signatures.iter().map(|sig_hex| {
+                        let sig_msg = clarity::util::secp256k1::MessageSignature::from_hex(sig_hex)
+                            .expect("unable to parse signer signature message");
+                        let pubkey = get_signer_pubkey_from_message_hash(&signer_sig_hash, &sig_msg)
+                            .expect("unable to recover signer sig pubkey");
+                        format!("0x{}", hex::encode(pubkey))
+                    }).collect())
+                }
+                _ => None,
+            },
```
zone117x, line 526: @rafaelcr any suggestions on how to add a test for this? I have a sample `/new_block` message and the expected pubkeys to use as a test vector, but not sure where/how to set that up.

components/chainhook-sdk/src/indexer/stacks/mod.rs
```diff
@@ -502,6 +512,19 @@ pub fn standardize_stacks_block(
             signer_bitvec: block.signer_bitvec.clone(),
             signer_signature: block.signer_signature.clone(),
 
+            signer_public_keys: match (signer_sig_hash, &block.signer_signature) {
+                (Some(signer_sig_hash), Some(signatures)) => {
+                    Some(signatures.iter().map(|sig_hex| {
+                        let sig_msg = clarity::util::secp256k1::MessageSignature::from_hex(sig_hex)
+                            .expect("unable to parse signer signature message");
+                        let pubkey = get_signer_pubkey_from_message_hash(&signer_sig_hash, &sig_msg)
+                            .expect("unable to recover signer sig pubkey");
```
rafaelcr, line 521: Can you `map_err` instead of `expect` to propagate the error upwards / avoid a panic if this goes wrong?

components/chainhook-sdk/src/indexer/stacks/mod.rs
rafaelcr: That's a good question, haven't quite figured out how to properly do tests yet, we should put a pin on this and I'll write some tests once we have the entire signers feature wired up and working

rafaelcr (approved):

PR #663, opened 2024-10-24
fix: update SignerMessage deserializing

zone117x:


components/chainhook-sdk/src/chainhooks/stacks/mod.rs
```diff
@@ -873,7 +873,11 @@ pub fn evaluate_stacks_predicate_on_non_consensus_events<'a>(
             | StacksPredicate::NftEvent(_)
             | StacksPredicate::StxEvent(_)
             | StacksPredicate::PrintEvent(_)
-            | StacksPredicate::Txid(_) => unreachable!(),
+            | StacksPredicate::Txid(_) => {
+                ctx.try_log(|logger| {
+                    slog::error!(logger, "Invalid predicate for non-consensus events: {:?}", chainhook.predicate)
+                });
```
rafaelcr, line 879: You should just return `false` here instead. I think part of chainhook's functionality is that it evaluates all predicates against all possible outcomes so this will be a common code path

rafaelcr (commented):
Thanks for the vscode config! will be very helpful

rafaelcr (approved):

PR #672, opened 2024-10-30
feat: Tenure change predicate

### LINKED ISSUE #522: Add predicates for `TenureChange` transactions
There's the potential for some new predicate types based off of the new `TenureChange` transactions.

If any users have requested types of predicates regarding the tenure change transactions, we'd love to hear about it here!

tippenein:
closes #522

### DIFF
--- a/components/chainhook-types-rs/src/events.rs
+++ b/components/chainhook-types-rs/src/events.rs
@@ -123,7 +123,13 @@ pub struct SmartContractEventData {
 
 #[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
 pub struct TenureChangeEventData {
-    pub consensus_hash: String,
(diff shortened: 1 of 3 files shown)


components/chainhook-types-rs/src/events.rs
```diff
@@ -121,6 +121,11 @@ pub struct SmartContractEventData {
     pub hex_value: String,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub struct TenureChangeEventData {
+    pub consensus_hash: String,
```
rafaelcr, line 126: We should include the fields from the `tenure_change_payload` example here:
https://api.hiro.so/extended/v1/tx/0x748962f013310e3276eea20cce032af876fe8dc2031367c4e46f5bfa2cd354c0

That should be enough

PR #675, opened 2024-10-31
feat: callback to intercept and modify predicates before re-registration

### LINKED ISSUE #674: PredicateHealthCheck re-register bug
The chainhook client healthcheck code detects inactive predicates, and re-registers using the same predicate from when the app was started.

This causes chainhook to re-send old events starting at the block_height that the client specified on startup.

```ts
async function startChainhookServer() {
  const blockHeight = await db.getChainTipBlockHeight();
  const predicate = {
    name: 'metadata-api-blocks',
    version: 1,
    chain: 'stacks',
    networks: {
      mainnet: {
        start_block: blockHeight, // <-- predicate re-registration uses whatever this value was on app start
        include_contract_abi: true,
        if_this: {
          scope: 'block_height',
          higher_than: 1,
        },
      },
    },
  };
  const observer: EventObserverOptions = {
    predicate_health_check_interval_ms: 300_000,
    // ...
  };
  const server = new ChainhookEventObserver(observer, { ... });
  await server.start([predicate], payload => db.processPayload(payload));
}
```

This could be solved with something like a callback where the client can provide an updated predicate for re-registration. E.g.:

```ts
  const observer: EventObserverOptions = {
    predicate_health_check_interval_ms: 300_000,
    predicate_reregister_callback: async (predicate) => {
      predicate.networks.mainnet.start_block = await db.getChainTipBlockHeight();
      return predicate;
    },
  };
```

zone117x:
Closes https://github.com/hirosystems/chainhook/issues/674

* First commit adds an optional callback that clients can use to modify an inactive predicate before it is re-registered.
* Second commit fixes an issue where the the `setInterval` could result in overlapping healthcheck code running if something like the callback (which could be waiting on a postgres connection, for example) take a long time.

### DIFF
--- a/components/client/typescript/src/index.ts
+++ b/components/client/typescript/src/index.ts
@@ -130,12 +130,22 @@ export class ChainhookEventObserver {
     this.fastify = await buildServer(this.observer, this.chainhook, predicates, callback);
     await this.fastify.listen({ host: this.observer.hostname, port: this.observer.port });
     if (this.observer.predicate_health_check_interval_ms && this.healthCheckTimer === undefined) {
-      this.healthCheckTimer = setInterval(() => {
-        predicateHealthCheck(this.observer, this.chainhook).catch(err =>
-          logger.error(err, `ChainhookEventObserver predicate health check error`)
-        );
-      }, this.observer.predicate_health_check_interval_ms);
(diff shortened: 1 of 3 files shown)


rafaelcr (approved):
LGTM

PR #676, opened 2024-10-31
Fix: async fs operations

zone117x:
Replaces the sync `node:fs` operations with async `node:fs/promises` async operations. This fixes potential issues with the main program loop completely blocking while waiting on disk operations. E.g. sync disk operations causing lots of small pauses/delays while serving http requests.


rafaelcr (approved):
Great!

PR #681, opened 2024-11-05
message_id should not be nullable

csgui:
`message_id` should not be nullable.

The tables `mock_proposals` and `mock_signatures` have an **UNIQUE** constraint on message_id, which suggests a one-to-one relationship between `messages` and `mock_proposals`, as well as between `messages` and `mock_signatures`. Therefore, `message_id` is used as a foreign key and should not allow **NULL** values.

### DIFF
--- a/components/chainhook-cli/src/storage/signers.rs
+++ b/components/chainhook-cli/src/storage/signers.rs
@@ -108,7 +108,7 @@ pub fn initialize_signers_db(base_dir: &PathBuf, ctx: &Context) -> Result<Connec
     conn.execute(
         "CREATE TABLE IF NOT EXISTS mock_proposals (
             id INTEGER PRIMARY KEY AUTOINCREMENT,
-            message_id INTEGER NOT NULL,
+            message_id INTEGER,
             burn_block_height INTEGER NOT NULL,
             stacks_tip_consensus_hash TEXT NOT NULL,
             stacks_tip TEXT NOT NULL,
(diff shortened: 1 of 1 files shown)


components/chainhook-cli/src/storage/signers.rs
```diff
@@ -143,7 +143,7 @@ pub fn initialize_signers_db(base_dir: &PathBuf, ctx: &Context) -> Result<Connec
         "CREATE TABLE IF NOT EXISTS mock_signatures (
             id INTEGER PRIMARY KEY AUTOINCREMENT,
             mock_proposal_id INTEGER NOT NULL,
-            message_id INTEGER,
+            message_id INTEGER NOT NULL,
```
zone117x, line 146: I think these need to be added using migration / alter-table logic otherwise they won't apply to existing dbs

csgui: @zone117x Are those migrations on `signers-metrics-api`?

zone117x: No the change in this PR updates this sqlite schema but those updates will only apply if starting with a fresh database. If this PR is ran against the current chainhook instance then the tables already exist and this change won't have any effect.

smcclellan:
@zone117x Is it possible for a mock proposal or mock signature to have no message (no corresponding entry in the `messages` table)?

csgui:
@rafaelcr Thanks for the info! 

I see the advantages of this approach in reducing schema complexity and avoiding duplicate table structures. However, in terms of data integrity—if a message is deleted (considering the CASCADE), we lose the peer info. But what if this peer info was actually related to a signature? Is this a real possibility?

rafaelcr:
@csgui not really. Messages write the data they need in that moment onto the DB, i.e. they don't look for repeated peer info so they can mark the relationship if it existed previously (it actually won't because it writes block data in the same row). 

If a message gets deleted and deletes peer info with it, no other data will be affected.

components/chainhook-cli/src/storage/signers.rs
csgui: Ok. Resolving this conversation and will change the implementation to a graceful error handling.

csgui:
Closing this PR, as this approach does not adequately address the problem. I will tackle the issue in a separate PR.

PR #682, opened 2024-11-06
propagate error when storing peer_info

csgui:
Adds error propagation when trying to insert `peer_info` data.

### DIFF
--- a/components/chainhook-cli/src/storage/signers.rs
+++ b/components/chainhook-cli/src/storage/signers.rs
@@ -49,7 +49,7 @@ pub fn initialize_signers_db(base_dir: &PathBuf, ctx: &Context) -> Result<Connec
     )
     .map_err(|e| format!("unable to create table: {e}"))?;
     conn.execute(
-        "CREATE INDEX IF NOT EXISTS index_messages_on_received_at ON messages(received_at_block_height)", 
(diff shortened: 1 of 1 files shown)


rafaelcr:
I was able to look at this briefly and I strongly suspect the problem is that we're being unable to write some of the `block_proposal` peer info data to the DB in the first place. This is why predicate evaluations fail when trying to retrieve those signer messages from the DB. @zone117x did you see any `unable to write mock proposal` log messages in your tests?

I noticed the problem could lie [here](https://github.com/hirosystems/chainhook/blob/stacks-signers/components/chainhook-cli/src/storage/signers.rs#L450), where we're not propagating a possible write error to the caller so we never notice the peer info wasn't written to the DB. (Oops, my bad)

@csgui could you add a `?` at the end of that statement and debug that before trying to silence these read errors? We might be able to fix the root of this problem. The PR as it stands right now will only mask the problem.

csgui:
@rafaelcr Thanks for the swift reply! Added the error propagation.

rafaelcr (approved):
Thanks @csgui ! We should try this out in mainnet and I'm pretty sure the DB write error will stick out pretty quickly

zone117x (approved):
Thanks @csgui!

PR #684, opened 2024-11-11
fix: only import stacks tsv if chainstate is empty

### LINKED ISSUE #677: Improvements to loading state from a TSV
Currently if the `tsv_file_url` property is configured, the chainhook node will attempt to download and restore from this file on reboot, even if there is pre-existing data (e.g. data restored from a volume snapshot).

Ideally the chainhook node would only download and restore from a TSV if its working dir is empty.

rafaelcr:
Also, remove the old code that attempted to re-import a TSV every 32 Stacks blocks

Fixes #677


zone117x (approved):
LGTM

PR #704, opened 2025-03-13
chore: clean dependencies

hugoclrd:
### Description

Remove some dependencies.
Mostly motivated by `prometheus` having an old version of `protobuf` in default features

Also updating the toolchain. I don't see any reason not to use the stable. (and 0.80 wasn't compiling)


rafaelcr (approved):
Thanks @hugocaillard

PR #706, opened 2025-03-25
chore: upgrade stacks-core

hugoclrd:
### Description

A recent PR in stacks-core updated the way cargo feature are handled in clarity and stacks-common (https://github.com/stacks-network/stacks-core/pull/5902).

Chainhook and Clarinet both need to update and handle this breaking changes - annoyingly, both repo need the other to do the update as well.

- this PR tgargets [this branch](https://github.com/hirosystems/clarinet/pull/1725) in clarinet, and would need to be merged first
- the clarinet PR will be updated once this one in merge to target chainhook/develop
- ideally, i will open an other PR on this repo to target a commit that is on clarinet/main

### CI changes

I think chainhook was built with older versions of rust and libclang out of extra caution.
But I don't think this is necessary, all of our other rust tools (including the blockchain itself) are built with stable.


rafaelcr (approved):

rafaelcr (approved):
LGTM

PR #733, opened 2025-11-14
ci: support npm trusted publishers

CharlieC3:
Related https://github.com/hirosystems/devops/issues/2230

* Adds NPM package to be managed and published by Semantic Release
* Adds support for [NPM trusted publishers](https://docs.npmjs.com/trusted-publishers)
* Updates GH actions

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -28,6 +28,20 @@
     [
       "@semantic-release/exec",
       {
+        "execCwd": "components/client/typescript",
+        "prepareCmd": "npm ci"
+      }
+    ],
+    [
+      "@semantic-release/npm",
+      {
+        "pkgRoot": "./components/client/typescript",
+        "npmPublish": true
+      }
+    ],
+    [
+      "@semantic-release/exec",
+      {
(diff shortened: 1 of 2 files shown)


.releaserc
```diff
@@ -25,6 +25,20 @@
         "preset": "conventionalcommits"
       }
     ],
+    [
+      "@semantic-release/exec",
+      {
+        "execCwd": "components/client/typescript",
+        "prepareCmd": "npm ci"
+      }
+    ],
```
CharlieC3, line 34: Does this package also need to be built to publish it? Not all packages do.

rafaelcr: This client is actually not integrated into semantic release yet and we've had to do the publishing by hand. We won't be able to integrate it into this file as easily because it has a completely different version than the repo release package.

We're deprecating this repo very soon so I think we should just keep it as is for now, what do you think?

CharlieC3: Keeping it as-is would be ok in that context. This effort is mostly to prevent CI from breaking from NPM removes all tokens in early December. If this release has not been integrated with CI, then there's nothing to break :)

Will close this one out.

